PR #9, opened 2026-03-26
Improve scheduler unit test coverage

DominicBurkart:
## Summary

The scheduler module (`src/scheduler.rs`) is the simulation engine driving all nodes and interferers — the most complex module at 290 lines of event-driven logic. It had only 6 unit tests, with several critical code paths at zero direct coverage:

- **Two-node collision recording** — the `if collided` branch in `deliver_completed_to_nodes`
- **Active interferer injection** — the entire `InterferencePoll` event handler
- **Airtime accumulation** — `record_airtime()` calls in both node and interferer paths
- **Post-delivery poll_transmit sweep** — the request-reply pattern (lines 212-220)
- **Broadcast to multiple receivers** — multi-node delivery loop
- **Capture effect routing** — `captured` flag check through the scheduler
- **Interferer-node collision** — interferer injections participating in the collision model

Adds 2 focused test helpers (`ActiveInterferer`, `ReplyNode`) and 8 new tests (7 deterministic + 1 proptest), bringing the scheduler from 6 to 14 unit tests. No existing tests modified.

### New tests

| Test | Invariant validated |
|------|-------------------|
| `two_nodes_overlapping_tx_records_collision` | Overlapping same-SF/freq TXs produce collisions, zero deliveries |
| `active_interferer_injects_and_records_airtime` | Interferer injections accumulate airtime but not `total_tx` |
| `single_tx_airtime_is_recorded` | Node TX records exact `duration_us` as airtime |
| `receive_triggers_reply_tx` | Request-reply works via post-delivery poll_transmit sweep |
| `broadcast_to_three_receivers` | Single TX delivered to all non-sender nodes |
| `capture_effect_through_scheduler` | Strong TX delivered, weak TX collided, both counters correct |
| `interferer_collides_with_node_tx` | Interferer injections participate in channel collision model |
| `n_receivers_all_get_broadcast` (proptest) | Broadcast delivery scales linearly with receiver count |

## Test plan

- [x] `cargo test --all-features` — all 85 tests pass (46 unit + 10 integration + 28 doc + 1 main)
- [x] `cargo clippy --all-features -- -D warnings` — clean
- [x] `cargo fmt --all --check` — formatted
- [ ] CI coverage report confirms increased project-level coverage

https://claude.ai/code/session_01DcmiJ6xZ5Tj9F3njmjPWE9

### DIFF
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -561,6 +561,17 @@ mod tests {
     }
 
     #[test]
+    fn add_node_after_interferer_validates_id_space() {
+        let mut sched = Scheduler::new(200_000);
+        sched.add_interferer(Box::new(NoOpInterferer), 0);
+        // Adding a node after an interferer exercises the debug_assert!
+        // closure that checks for NodeId/interferer ID space collisions.
+        sched.add_node(Box::new(SimpleNode::new(1)), Some(0));
+        sched.run();
+        assert_eq!(sched.metrics.total_tx, 0);
+    }
+
+    #[test]
     fn two_nodes_overlapping_tx_records_collision() {
         let mut sched = Scheduler::new(200_000);
         let mut n1 = SimpleNode::new(1);
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -466,9 +466,7 @@ mod tests {
             self.id
         }
         fn on_receive(&mut self, _f: RxMetadata, _t: SimTime) -> Option<SimTime> {
-            if !self.received {
-                self.received = true;
-            }
+            self.received = true;
             None
         }
         fn poll_transmit(&mut self, _t: SimTime) -> Option<Transmission> {
@@ -583,9 +581,9 @@ mod tests {
         sched.run();
 
         assert_eq!(sched.metrics.total_tx, 2);
-        assert!(
-            sched.metrics.total_collisions > 0,
-            "overlapping same-SF/freq TXs must collide"
+        assert_eq!(
+            sched.metrics.total_collisions, 2,
+            "overlapping same-SF/freq TXs must both collide"
         );
         assert_eq!(sched.metrics.total_rx, 0);
     }
@@ -692,7 +690,10 @@ mod tests {
         assert_eq!(sched.metrics.total_tx, 2);
         assert_eq!(sched.metrics.total_captures, 1);
         assert_eq!(sched.metrics.total_collisions, 1);
-        assert!(sched.metrics.total_rx > 0, "strong TX must be delivered");
+        assert_eq!(
+            sched.metrics.total_rx, 2,
+            "strong TX delivered to both non-sender nodes"
+        );
     }
 
     #[test]
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -318,6 +318,7 @@ impl Scheduler {
 mod tests {
     use super::*;
     use crate::types::{ChannelEvent, Transmission};
+    use proptest::prelude::*;
 
     struct SimpleNode {
         id: NodeId,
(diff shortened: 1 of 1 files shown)


src/scheduler.rs
```diff
@@ -430,6 +430,59 @@ mod tests {
...
+        reply_tx: Option<Transmission>,
+        received: bool,
+    }
+
+    impl NodeHandle for ReplyNode {
+        fn node_id(&self) -> NodeId {
+            self.id
+        }
+        fn on_receive(&mut self, _f: RxMetadata, _t: SimTime) -> Option<SimTime> {
+            if !self.received {
```
DominicBurkart: Nit: The `if !self.received` guard is redundant — assigning `self.received = true` is already idempotent. The guard makes it look like there's a reason to skip on subsequent receives (e.g. a side effect), but there isn't. Simplifying to just `self.received = true;` would be clearer.

```suggestion
            self.received = true;
```

---

src/scheduler.rs
```diff
@@ -506,4 +559,164 @@ mod tests {
...
+    fn two_nodes_overlapping_tx_records_collision() {
+        let mut sched = Scheduler::new(200_000);
+        let mut n1 = SimpleNode::new(1);
+        n1.queue_tx(make_tx(7, 868_100_000, 50_000));
+        let mut n2 = SimpleNode::new(2);
+        n2.queue_tx(make_tx(7, 868_100_000, 50_000));
+        sched.add_node(Box::new(n1), Some(0));
+        sched.add_node(Box::new(n2), Some(10_000));
+        sched.run();
+
```
DominicBurkart, line 582: This assertion could be tightened. Two overlapping same-SF/freq TXs at equal power (both 14 dBm from `make_tx`) will each be marked collided when resolved — so `total_collisions` should be exactly 2. Using `assert_eq!` here would catch a regression where, say, only one side of a collision is recorded.

```suggestion
        assert_eq!(
            sched.metrics.total_collisions, 2,
            "overlapping same-SF/freq TXs must both collide"
        );
```

---

src/scheduler.rs
```diff
@@ -506,4 +559,164 @@ mod tests {
...
+        fn n_receivers_all_get_broadcast(n in 2usize..20) {
+            let mut sched = Scheduler::new(200_000);
+            let mut sender = SimpleNode::new(0);
+            sender.queue_tx(make_tx(7, 868_100_000, 50_000));
+            sched.add_node(Box::new(sender), Some(0));
+            for i in 1..=n {
+                sched.add_node(Box::new(SimpleNode::new(i as u32)), None);
+            }
+            sched.run();
+            prop_assert_eq!(sched.metrics.total_tx, 1u64);
```
DominicBurkart, line 730: This assertion is weaker than it needs to be. With the strong TX (node 1, 20 dBm) successfully delivered to both non-sender nodes (2 and 3), `total_rx` should be exactly 1 — wait, actually it should be **2**: node 2 (the weak sender) and node 3 (the passive receiver) both receive the strong TX. Consider:

```suggestion
        assert_eq!(sched.metrics.total_rx, 2, "strong TX delivered to both non-sender nodes");
```

This would also catch a regression where the capture-effect path accidentally skips delivery to one receiver.

---

DominicBurkart (commented):
Nice work — this is a solid set of tests targeting meaningful coverage gaps in the scheduler. The test helpers (`ActiveInterferer`, `ReplyNode`) are well-scoped, the production code cleanup (`sender: _`) is clean, and the proptest for broadcast scaling is a good addition.

**Summary of inline comments:**

1. **`ReplyNode::on_receive`** — the `if !self.received` guard is redundant since the assignment is idempotent.
2. **`two_nodes_overlapping_tx_records_collision`** — `total_collisions > 0` could be `== 2` for a stronger regression check (both TXs are resolved as collided).
3. **`capture_effect_through_scheduler`** — `total_rx > 0` could be `== 2` (the strong TX is delivered to both non-sender nodes 2 and 3).

**One more note** (couldn't attach inline — outside diff hunk):

- In `interferer_collides_with_node_tx`, `poll_interval: 0` works because `remaining: 1` ensures only one injection, but it reads as "poll again immediately." A non-zero interval (e.g. `50_000`) would make the single-injection intent clearer and avoid a subtle same-timestamp footgun if someone later bumps `remaining`.

---

PR #14, opened 2026-03-27
Add tests for undertested core components

DominicBurkart:
## Summary\n\nAdds `tests/core_coverage.rs` with 25+ tests (including proptests) targeting gaps in the existing test suite:\n\n- **Channel edge cases**: empty resolve, `resolve_at` before TX ends, `drain_completed` idempotency, `deliver_to` non-destructiveness, partial resolve leaving active TXs, `begin_transmission` return event validation\n- **RSSI/SNR computation**: direct unit tests for `compute_rssi` and `compute_snr`, plus proptest for monotonicity\n- **Capture-effect edge cases**: weaker-arrives-first ordering, high co-channel rejection threshold\n- **Metrics per-node isolation**: independent TX/RX counts across nodes, TX vs RX non-interference, proptests asserting per-node sums equal totals\n- **Scheduler event ordering**: `EventKind` equality and variant distinctness\n- **Scheduler multi-node interleaving**: two periodic nodes on different frequencies, current_time advancement, empty scheduler termination, inert-node-still-receives\n- **Proptests**: solo TX never collides, resolve at exact end time always completes, RSSI/SNR monotonicity, metrics summation invariants\n\n## Rationale\n\nExisting tests thoroughly cover collision semantics and the LoRaWAN integration path, but leave gaps around:\n1. Channel methods called on empty/partially-resolved state\n2. The public `compute_rssi`/`compute_snr` API (only tested indirectly)\n3. Metrics per-node bookkeeping (only basic increment tests existed)\n4. Scheduler behavior with zero nodes, zero events, or multiple periodic nodes\n5. Property-based invariants for the metrics and channel subsystems\n\n## Test plan\n\n- [ ] CI passes `cargo test --all-features`\n- [ ] No clippy warnings introduced\n- [ ] Coverage delta is positive for `channel.rs`, `metrics.rs`, and `scheduler.rs`"

### DIFF
--- a/tests/core_coverage.rs
+++ b/tests/core_coverage.rs
@@ -31,13 +31,14 @@ fn default_tx() -> Transmission {
     make_tx(7, 868_100_000, 50_000, 14)
 }
 
-/// A node that transmits once per wake, scheduling the next wake after a
-/// configurable interval. Tracks how many times it was woken.
+/// A node that transmits once per wake (not on receive-triggered polls).
+/// Tracks how many times it was woken.
 struct CountingNode {
     id: NodeId,
     interval_us: u64,
     wake_count: u32,
-    tx: Option<Transmission>,
+    tx_template: Option<Transmission>,
+    ready_to_tx: bool,
     received: Vec<RxMetadata>,
 }
 
@@ -47,7 +48,8 @@ impl CountingNode {
             id: NodeId(id),
             interval_us,
             wake_count: 0,
-            tx,
+            tx_template: tx,
+            ready_to_tx: false,
             received: Vec::new(),
         }
     }
@@ -62,10 +64,16 @@ impl NodeHandle for CountingNode {
         None
     }
     fn poll_transmit(&mut self, _time: SimTime) -> Option<Transmission> {
-        self.tx.clone()
+        if self.ready_to_tx {
+            self.ready_to_tx = false;
+            self.tx_template.clone()
+        } else {
+            None
+        }
     }
     fn update(&mut self, time: SimTime) -> Option<SimTime> {
         self.wake_count += 1;
+        self.ready_to_tx = true;
         Some(time + self.interval_us)
     }
 }
@@ -90,6 +98,43 @@ impl NodeHandle for InertNode {
     }
 }
 
+/// A node that transmits exactly once on its first wake.
+struct OneShotNode {
+    id: NodeId,
+    tx: Option<Transmission>,
+    fired: bool,
+}
+
+impl OneShotNode {
+    fn new(id: u32, tx: Transmission) -> Self {
+        Self {
+            id: NodeId(id),
+            tx: Some(tx),
+            fired: false,
+        }
+    }
+}
+
+impl NodeHandle for OneShotNode {
+    fn node_id(&self) -> NodeId {
+        self.id
+    }
+    fn on_receive(&mut self, _f: RxMetadata, _t: SimTime) -> Option<SimTime> {
+        None
+    }
+    fn poll_transmit(&mut self, _time: SimTime) -> Option<Transmission> {
+        if self.fired {
+            self.tx.take()
+        } else {
+            None
+        }
+    }
+    fn update(&mut self, _time: SimTime) -> Option<SimTime> {
+        self.fired = true;
+        None // only one wake
+    }
+}
+
 // ===========================================================================
 // Channel edge-case tests
 // ===========================================================================
@@ -99,7 +144,6 @@ fn resolve_before_any_tx_ends_returns_empty() {
     let mut ch = Channel::new();
     let tx = default_tx(); // duration 50_000
     ch.begin_transmission(NodeId(1), &tx, 0);
-    // Resolve at a time before the TX ends
     let events = ch.resolve_at(25_000);
     assert!(events.is_empty(), "no TX should complete before its end time");
 }
@@ -332,7 +376,6 @@ fn two_periodic_nodes_both_transmit() {
 
     // Node 1: wakes at 0, 100k, 200k, 300k, 400k = 5 TXs
     // Node 2: wakes at 0, 150k, 300k, 450k = 4 TXs
-    // Each TX is received by the other node (different freq, no collision)
     assert_eq!(sched.metrics.node_tx_count(NodeId(1)), 5);
     assert_eq!(sched.metrics.node_tx_count(NodeId(2)), 4);
     assert_eq!(sched.metrics.total_tx, 9);
@@ -372,8 +415,8 @@ fn scheduler_with_no_events_terminates() {
 #[test]
 fn node_without_tx_still_receives() {
     let mut sched = Scheduler::new(200_000);
-    // Sender transmits once
-    let sender = CountingNode::new(1, 999_999_999, Some(default_tx()));
+    // Sender transmits exactly once
+    let sender = OneShotNode::new(1, default_tx());
     let listener = InertNode { id: NodeId(2) };
     sched.add_node(Box::new(sender), Some(0));
     sched.add_node(Box::new(listener), None);
@@ -390,8 +433,6 @@ fn node_without_tx_still_receives() {
 
 #[test]
 fn capture_weaker_arrives_first() {
-    // The weaker signal starts first, then the stronger one arrives.
-    // The stronger should capture (survive), weaker should be collided.
     let mut ch = Channel::new();
     let weak = make_tx(7, 868_100_000, 50_000, 8);
     let strong = make_tx(7, 868_100_000, 50_000, 20);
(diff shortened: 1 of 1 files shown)


tests/core_coverage.rs
```diff
@@ -0,0 +1,607 @@
...
+    }
+    fn on_receive(&mut self, _f: RxMetadata, _t: SimTime) -> Option<SimTime> {
+        None
+    }
+    fn poll_transmit(&mut self, _time: SimTime) -> Option<Transmission> {
+        if self.fired {
+            self.tx.take()
+        } else {
+            None
+        }
```
DominicBurkart, line 130: `CountingNode` has a `wake_count` field and a `received` vec that are written to but never read in any test. `InertNode` and `OneShotNode` are clean, but `CountingNode` carries dead bookkeeping weight.

Two options:
1. Drop the unused fields to keep the helper minimal.
2. Add assertions that actually use them (e.g. assert `wake_count` in `two_periodic_nodes_both_transmit`) -- that would make the test independent of `metrics` correctness, which is arguably better.

---

tests/core_coverage.rs
```diff
@@ -0,0 +1,607 @@
...
+    let first = ch.deliver_to(50_000);
+    assert_eq!(first.len(), 1);
+
+    // deliver_to is non-destructive; calling again should yield the same result
+    let second = ch.deliver_to(50_000);
+    assert_eq!(second.len(), 1);
+}
+
+#[test]
+fn compute_rssi_reflects_tx_power() {
```
DominicBurkart, line 195: These assertions pin the exact formula (`rssi = tx_power - 100`, `snr = rssi + 117`). That is correct today, but it means this test will break if `Channel` ever gains distance-based path loss or a configurable noise floor at construction time.

Consider whether the behavioral assertion alone (the `rssi_high > rssi_low` check above) is sufficient for the unit test, and move the exact-value checks into a clearly-named test like `default_channel_parameters` so the coupling to implementation constants is explicit and easy to find when those constants change.

---

tests/core_coverage.rs
```diff
@@ -0,0 +1,607 @@
...
+    assert_eq!(m.total_rx, 1);
+}
+
+// ===========================================================================
+// Scheduler event ordering tests
+// ===========================================================================
+
+#[test]
+fn event_kind_wake_equality() {
+    let a = EventKind::Wake {
```
DominicBurkart, line 324: The `EventKind` equality tests (`event_kind_wake_equality`, `event_kind_tx_complete_equality`, `event_kind_variants_are_distinct`) are testing a derived `PartialEq` impl rather than any behavioral invariant of the simulator. These will pass for any correct `#[derive(PartialEq)]` and will never catch a real bug.

I would drop these three tests. If the concern is that someone removes the derive, the compiler will catch that the moment `ScheduledEvent`'s `PartialEq` (or any match on `EventKind`) fails to compile.

---

tests/core_coverage.rs
```diff
@@ -0,0 +1,607 @@
...
+    assert_ne!(wake, poll);
+    assert_ne!(tx, poll);
+}
+
+// ===========================================================================
+// Scheduler: multi-node interleaving
+// ===========================================================================
+
+#[test]
+fn two_periodic_nodes_both_transmit() {
```
DominicBurkart, line 367: The inline comment says `Node 1: wakes at 0, 100k, 200k, 300k, 400k, 500k = 6 TXs` and asserts `node_tx_count(NodeId(1)) == 6`. This depends on the scheduler processing events where `time <= end_time` (inclusive). If the scheduler loop condition ever changes to `time < end_time`, this test breaks with a confusing off-by-one. That is fine as a regression test, but the comment should be more explicit: "Asserts the scheduler uses `>` (not `>=`) to break, so t=500_000 == end_time is processed."

Also: this test validates the same `two_periodic_nodes_both_transmit` scenario as the existing `simulation_stops_at_end_time` test in `src/scheduler.rs` (which already covers the inclusive-boundary behavior). The added value here is the per-node TX count check, which is good -- just calling it out so readers understand the incremental coverage.

---

tests/core_coverage.rs
```diff
@@ -0,0 +1,607 @@
...
+    sched.add_node(Box::new(listener), None);
+    sched.run();
+
+    assert_eq!(sched.metrics.total_tx, 1);
+    assert_eq!(sched.metrics.node_rx_count(NodeId(2)), 1);
+    assert_eq!(sched.metrics.node_rx_count(NodeId(1)), 0);
+}
+
+// ===========================================================================
+// Channel: capture-effect edge cases
```
DominicBurkart, line 434: Nice test. One thing to verify: the assertion `strong_entry.2` (captured == true) relies on the tuple shape `(NodeId, bool, bool, RxMetadata)` from `drain_completed`. If that return type changes (e.g. to a struct), this breaks silently at compile time but the field semantics could drift. A brief comment like `// .1 = collided, .2 = captured` would help future readers.

This applies to `collided_tx_not_in_deliver_to` as well (line 218), where `completed.iter().all(|(_, collided, _, _)| *collided)` is indexing into the same tuple.

---

tests/core_coverage.rs
```diff
@@ -0,0 +1,607 @@
...
+    ch.begin_transmission(NodeId(1), &weak, 0);
+    ch.begin_transmission(NodeId(2), &strong, 10_000);
+    ch.resolve_at(60_000);
+
+    let completed = ch.drain_completed();
+    let weak_entry = completed
+        .iter()
+        .find(|(id, _, _, _)| *id == NodeId(1))
+        .unwrap();
+    let strong_entry = completed
```
DominicBurkart, line 451: The assertion message says `"delta=30 < threshold=50, both should collide"` but the delta is actually `20 - (-10) = 30` dB of TX power difference. The *RSSI* delta is also 30 since `compute_rssi` is a simple subtraction. So the comment is correct, but it is worth noting that this test implicitly depends on the fact that path loss cancels out in the delta computation (both signals experience the same 100 dB loss). If the channel ever gains per-link path loss, this test's reasoning breaks even though the assertion might still pass by coincidence.

Consider adding a comment to that effect, or testing with `compute_rssi` values directly to make the intention clearer.

---

tests/core_coverage.rs
```diff
@@ -0,0 +1,607 @@
...
+
+// ===========================================================================
+// Proptests
+// ===========================================================================
+
+#[cfg(test)]
+mod proptests {
+    use super::*;
+    use proptest::prelude::*;
+
```
DominicBurkart, line 510: The `rssi_monotone_with_tx_power` and `snr_monotone_with_rssi` proptests are testing that simple arithmetic (`x - constant`) is monotone. This will always hold for subtraction by a constant, so the proptest machinery (shrinking, case generation) does not buy anything over the unit tests above.

These are not harmful, but they are low-signal. If you want to keep them, consider combining them into a single property that tests the full pipeline: given two TX powers `a <= b`, assert `snr(rssi(a)) <= snr(rssi(b))` -- that at least exercises composition.

---

tests/core_coverage.rs
DominicBurkart, line 1: **Missing coverage to consider as follow-ups:**

The now-closed PR #15 contained several test scenarios that are not covered by this PR or the existing test suite and would be valuable additions:

1. **Multi-interferer collision/orthogonality** -- two `InterferenceSource` instances on the same SF/freq collide; on different SFs they do not. The existing test suite only exercises a single interferer at a time.
2. **Multi-SF concurrent transmissions through the full scheduler path** -- 6 nodes on SF 7-12 all transmit simultaneously and all deliver without collision. SF orthogonality is currently tested at the `Channel` level but not through scheduler dispatch.
3. **Back-to-back TX through the scheduler** -- a node starts transmitting at exactly the time another finishes (`t=50_000`). This exercises the `overlaps()` boundary through the full event loop, not just the channel unit test.
4. **Mixed node + interferer airtime accounting** -- asserting `total_airtime_us` sums contributions from both nodes and interferers.
5. **`drain_completed` then `deliver_to` returns empty** -- the inverse of the `deliver_to_does_not_drain` test in this PR. Important for verifying the interaction semantics between the two methods.

These would round out the coverage nicely and could be a fast follow-up PR.

---

DominicBurkart (commented):
Good set of tests that fills real gaps in the existing suite. The channel edge-case tests (empty resolve, partial resolve, idempotent drain vs non-destructive deliver) and the scheduler termination/interleaving tests are the highest-value additions here. The capture-effect ordering test (`capture_weaker_arrives_first`) is a nice complement to the existing `stronger_signal_survives_capture` test which only covers the strong-first ordering.

**Summary of feedback:**

- **Drop the `EventKind` equality tests** (lines 313-324). They test a derived `PartialEq` impl, not simulator behavior. The compiler already enforces this.
- **Trim or use `CountingNode`'s dead fields** (`wake_count`, `received`). Either assert on them to make tests independent of `MetricsCollector`, or remove them to keep the helper minimal.
- **The RSSI/SNR exact-value assertions** couple tests to `Channel::new()` default constants. Consider isolating those into a clearly-named test so the coupling is intentional and easy to update.
- **The RSSI/SNR monotonicity proptests** are testing trivially-true arithmetic properties. Low harm, but low signal -- consider collapsing into a single composed-pipeline property.
- **Add tuple field comments** where tests destructure `CompletedTx` tuples (`.1 = collided`, `.2 = captured`) to guard against silent semantic drift.

**Follow-up coverage from closed PR #15** worth adding: multi-interferer collision/orthogonality, multi-SF delivery through the scheduler, back-to-back TX through the event loop, mixed airtime accounting, and `drain_completed` then `deliver_to` interaction. See file-level comment for details.

---

PR #24, opened 2026-04-01
feat: parameterize channel model for non-LoRa protocol support

DominicBurkart:
## Summary

- Introduces `ChannelConfig` — a plain struct holding the three physical-layer parameters (`path_loss_db`, `noise_floor_dbm`, `co_channel_rejection_db`) with named constants for the LoRa defaults and a `lora_defaults()` constructor.
- Adds `Channel::with_config(config: ChannelConfig) -> Channel` as the new primary entry point for non-LoRa protocols; `Channel::new()` and `Channel::with_co_channel_rejection()` are preserved unchanged for full backward compatibility.
- Adds `Scheduler::with_channel(end_time, channel)` so end-to-end simulations can inject a custom channel without touching internal state.
- Updates all internal field accesses that previously used bare `self.path_loss_db` / `self.noise_floor_dbm` / `self.co_channel_rejection_db` to go through `self.config.*`.
- Adds `Channel::config() -> &ChannelConfig` accessor.
- Adds 4 new unit tests in `channel.rs` covering: `ChannelConfig` default/constants round-trip, `config()` accessor, lower path loss → higher RSSI, lower noise floor → higher SNR, and higher capture threshold → prevents capture where LoRa would survive.
- Adds 1 new integration-level test in `scheduler.rs` (`with_channel_uses_custom_config`) showing that the strict channel produces different collision outcomes than the LoRa default.

## Design notes

- `ChannelConfig` is a plain `#[derive(Debug, Clone, PartialEq)]` struct with `pub` fields — no builder boilerplate. This matches the existing codebase style (no builder pattern anywhere), keeps it easy to construct with struct literal syntax, and allows field-update syntax (`..ChannelConfig::lora_defaults()`).
- All fields stay `f32` to match the existing codebase convention (no silent widening to `f64`).
- `ChannelConfig` implements `Default` (delegates to `lora_defaults()`) so it composes naturally with `..Default::default()` patterns.
- The old `with_co_channel_rejection` constructor now delegates to `with_config`, eliminating the previous partial duplication.

## Test plan

- [ ] `cargo test` — all existing unit and integration tests pass unmodified
- [ ] `cargo test --doc` — all new doctests compile and pass
- [ ] Verify `Channel::new()` still works as before (backward compatible)
- [ ] Verify `Scheduler::new()` still works as before (backward compatible)
- [ ] Run the three new `channel.rs` parametric tests manually and confirm RSSI/SNR assertions hold
- [ ] Confirm `with_channel_uses_custom_config` in `scheduler.rs` produces `total_rx=2` (LoRa) vs `total_rx=0` (strict threshold)

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -3,6 +3,82 @@ use crate::types::{ChannelEvent, NodeId, RxMetadata, Transmission};
 
 pub type CompletedTx = (NodeId, bool, bool, RxMetadata);
 
+/// Default LoRa path loss in dB (free-space + typical indoor attenuation baseline).
+pub const LORA_PATH_LOSS_DB: f32 = 100.0;
+
+/// Default LoRa noise floor in dBm (LoRa sensitivity at SF7/125 kHz).
+pub const LORA_NOISE_FLOOR_DBM: f32 = -117.0;
+
+/// Default co-channel rejection threshold in dB (LoRa capture effect threshold).
+pub const LORA_CO_CHANNEL_REJECTION_DB: f32 = 6.0;
+
+/// Configuration for physical-layer channel parameters.
+///
+/// Construct with [`ChannelConfig::lora_defaults()`] for LoRa or supply
+/// custom values for other protocols.
+///
+/// # Examples
+///
+/// ```
+/// use theatron::channel::ChannelConfig;
+///
+/// // LoRa defaults
+/// let cfg = ChannelConfig::lora_defaults();
+///
+/// // Custom 802.15.4 / Zigbee-like parameters
+/// let cfg = ChannelConfig {
+///     path_loss_db: 80.0,
+///     noise_floor_dbm: -100.0,
+///     co_channel_rejection_db: 3.0,
+/// };
+/// ```
+#[derive(Debug, Clone, PartialEq)]
+pub struct ChannelConfig {
+    /// Path loss applied to every transmission (dB). Higher values mean
+    /// the received signal is weaker.
+    pub path_loss_db: f32,
+
+    /// Noise floor of the receiver (dBm). Used to compute SNR.
+    pub noise_floor_dbm: f32,
+
+    /// Minimum power difference (dB) required for the stronger signal to
+    /// survive a co-channel collision via the capture effect.
+    pub co_channel_rejection_db: f32,
+}
+
+impl ChannelConfig {
+    /// Return the LoRa default parameters used by [`Channel::new()`].
+    ///
+    /// | Parameter               | Value    |
+    /// |-------------------------|----------|
+    /// | `path_loss_db`          | 100 dB   |
+    /// | `noise_floor_dbm`       | -117 dBm |
+    /// | `co_channel_rejection_db` | 6 dB   |
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use theatron::channel::ChannelConfig;
+    /// let cfg = ChannelConfig::lora_defaults();
+    /// assert_eq!(cfg.path_loss_db, 100.0);
+    /// assert_eq!(cfg.noise_floor_dbm, -117.0);
+    /// assert_eq!(cfg.co_channel_rejection_db, 6.0);
+    /// ```
+    pub fn lora_defaults() -> Self {
+        Self {
+            path_loss_db: LORA_PATH_LOSS_DB,
+            noise_floor_dbm: LORA_NOISE_FLOOR_DBM,
+            co_channel_rejection_db: LORA_CO_CHANNEL_REJECTION_DB,
+        }
+    }
+}
+
+impl Default for ChannelConfig {
+    fn default() -> Self {
+        Self::lora_defaults()
+    }
+}
+
 struct ActiveTransmission {
     sender: NodeId,
     payload: Vec<u8>,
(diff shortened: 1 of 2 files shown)


DominicBurkart (commented):
## Code Review — feat: parameterize channel model for non-LoRa protocol support

**CI:** All 6 checks pass (Test, Check, Autoformat, Code Coverage, Cargo Deny, Security Audit). Coverage moves from 99.49% to 100.00%. No secrets detected.

---

### Summary verdict

This is a well-executed refactor. The approach is sound, backward compatibility is preserved, and the test suite is thorough. A few issues are worth addressing before merge, ranging from one correctness concern to several quality items.

---

### Issues found

#### 1. `ChannelConfig` fields are fully public — struct-literal construction is a semver hazard (medium)

`ChannelConfig` is `pub` with all three fields `pub`. That means any downstream crate that constructs it with a struct literal (`ChannelConfig { path_loss_db: …, noise_floor_dbm: …, co_channel_rejection_db: … }`) will get a **compile error** the moment a fourth field is added. The PR description explicitly calls out field-update syntax (`..ChannelConfig::lora_defaults()`) as a feature, but that syntax also breaks if a new non-`Default` field is ever added.

The existing codebase has no builder pattern, which is fine, but the standard Rust idiom for forward-compatible public structs with all-public fields is to add a `#[non_exhaustive]` attribute. This is a one-line change:

```rust
#[non_exhaustive]
#[derive(Debug, Clone, PartialEq)]
pub struct ChannelConfig { … }
```

`#[non_exhaustive]` prevents external struct-literal construction while still allowing the `lora_defaults()` / field-update pattern _within_ this crate. The only downside is that external callers must go through `lora_defaults()` + field-update syntax, which the PR already documents as the preferred approach.

If the intent is to keep struct-literal construction as a supported public API indefinitely, that is a reasonable call, but it should be a conscious decision noted in the PR description rather than an implicit one.

#### 2. `ChannelConfig` does not derive `Copy` (minor, but a correctness papercut)

All three fields are `f32`, which is `Copy`. The struct is small (12 bytes). Not deriving `Copy` means callers must `.clone()` in trivial contexts — the new test `with_config_exposes_config_accessor` already does `cfg.clone()` to work around this. Adding `Copy` to the derive list is zero-cost and removes friction.

```rust
#[derive(Debug, Clone, Copy, PartialEq)]
pub struct ChannelConfig { … }
```

#### 3. Named constants are correct but the doc comments contain a minor inaccuracy (minor)

`LORA_PATH_LOSS_DB` is documented as "free-space + typical indoor attenuation baseline." Free-space path loss at 868 MHz over a typical indoor LoRaWAN link distance is well under 100 dB; 100 dB is more representative of a moderate urban/suburban outdoor link budget. The comment is not wrong enough to cause a bug, but it may mislead users choosing parameters for non-LoRa protocols. Consider replacing it with a neutral description such as "conservative link-budget baseline used as the LoRa simulation default" or citing the actual derivation.

#### 4. `Scheduler::with_channel` takes `Channel` by value — no way to inspect or replace the channel after construction (design note, not a blocker)

The new `with_channel` constructor takes `Channel` by value, which is correct for initialization. However, `Scheduler` has no `channel()` accessor, so it is impossible for callers to verify which `ChannelConfig` is active in a running simulation. The PR adds `Channel::config() -> &ChannelConfig`, which is good, but that accessor is unreachable once the channel is moved into `Scheduler`.

This is not a blocking issue for the PR's stated goals, but a `scheduler.channel_config() -> &ChannelConfig` pass-through would complete the ergonomic picture and is trivially implementable.

#### 5. `with_channel_uses_custom_config` integration test contains unnecessary duplication (nit)

The scheduler integration test in `scheduler.rs` builds two near-identical `Transmission` structs twice (once for `sched_lora`, once for `sched_strict`). Since `Transmission` presumably implements `Clone` (or can), the two node setups could share a helper. This is a style nit with no correctness impact, but it doubles the visual noise and increases the maintenance surface when `Transmission` fields change. A local `make_tx_power` helper (as already used in `channel.rs` tests) would clean this up considerably.

#### 6. Pending Kani coverage gap (informational)

The open issues #32 and #33 call for Kani proofs of the capture-effect boundary conditions and collision detection in `channel.rs`. This PR changes the capture-effect threshold from a hardcoded field to `self.config.co_channel_rejection_db`. The logic itself is unchanged, so the existing proof targets are still valid, but whoever picks up #32/#33 should note that the threshold is now accessed via the config struct — the harness setup will need a `ChannelConfig` literal rather than a direct field assignment.

---

### What is correct and well done

- **Backward compatibility:** `Channel::new()`, `Channel::with_co_channel_rejection()`, and `Scheduler::new()` all delegate to the new constructors without any behavior change. The old construction paths are fully preserved.
- **Named constants:** `LORA_PATH_LOSS_DB`, `LORA_NOISE_FLOOR_DBM`, `LORA_CO_CHANNEL_REJECTION_DB` match the hardcoded values that were in `Channel::new()` on `main` (100.0, -117.0, 6.0 respectively). The round-trip test `channel_config_lora_defaults_matches_constants` confirms this.
- **`Default` delegation:** `ChannelConfig::default()` correctly delegates to `lora_defaults()`, consistent with the rest of the codebase where `Channel::default()` delegates to `Channel::new()`.
- **`with_co_channel_rejection` refactor:** Switching from `..Self::new()` to `..ChannelConfig::lora_defaults()` is a genuine improvement — it eliminates an implicit dependency on `Channel::new()`'s struct state and makes the defaults explicit.
- **Test coverage:** The four new `channel.rs` tests cover the constant round-trip, accessor correctness, RSSI ordering, SNR ordering, and capture-threshold polarity. The scheduler integration test exercises the full end-to-end path. Coverage is at 100%.
- **No secrets or sensitive values in the diff.**
- **`ChannelConfig` is not exported from `lib.rs` at the crate root** — it is accessible as `theatron::channel::ChannelConfig`, which is consistent with how `Channel` and `CompletedTx` are exposed. No re-export churn.


---

PR #27, opened 2026-04-03
test: Protocol trait contract tests

DominicBurkart:
Adds a minimal conforming Protocol implementation and three tests validating the trait contract: init wake time, poll_transmit idempotence, and on_receive safety.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,6 +1,9 @@
 coverage:
   status:
-    project: off
-  patch:
-    default:
-      target: 100%
+    project:
+      default:
+        target: auto
+        threshold: 0%
+    patch:
+      default:
+        target: 100%
--- a/tests/protocol_contract.rs
+++ b/tests/protocol_contract.rs
@@ -1,0 +1,109 @@
+//! Contract tests for the Protocol trait.
+//!
+//! These tests verify that a conforming implementation satisfies the
+//! behavioural expectations of the trait: init returns a wake time,
+//! poll_transmit drains after the first call, and on_receive is safe
+//! to call at any time.
+
+use theatron::traits::Protocol;
+use theatron::types::{RxMetadata, Transmission};
+
+/// The simplest possible Protocol implementation: transmits one fixed
+/// frame on the first poll, then goes silent.
+struct MinimalProtocol;
+
+/// Internal state: tracks whether the single transmission has been sent.
+struct MinimalState {
+    transmitted: bool,
+}
+
+impl Protocol for MinimalProtocol {
+    type Config = ();
+    type State = MinimalState;
+    type Metrics = ();
+
+    fn init(&self, _config: ()) -> (MinimalState, Option<u64>) {
+        (MinimalState { transmitted: false }, Some(0))
+    }
+
+    fn on_receive(
+        &self,
+        _state: &mut MinimalState,
+        _frame: RxMetadata,
+        _time: u64,
+    ) -> Option<u64> {
+        None
+    }
+
+    fn poll_transmit(
+        &self,
+        state: &mut MinimalState,
+        _time: u64,
+    ) -> Option<Transmission> {
+        if state.transmitted {
+            return None;
+        }
+        state.transmitted = true;
+        Some(Transmission {
+            payload: vec![0x01],
+            sf: 7,
+            bandwidth: 125_000,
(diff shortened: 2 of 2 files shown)


DominicBurkart (commented):
## Review — Protocol trait contract tests

CI is green (Test, Check, Autoformat, Security Audit, Cargo Deny all pass). No secrets or PII. The structure is clean. However there are several correctness and completeness issues that need to be addressed before this merges.

---

### [MUST FIX] `metrics()` implementation silently omits the return type

`Protocol::metrics` is declared as `fn metrics(&self, state: &Self::State) -> Self::Metrics`. The stub implementation writes:

```rust
fn metrics(&self, _state: &MinimalState) {}
```

This compiles only because `type Metrics = ()` and `{}` implicitly returns `()`. The implementation is missing `-> Self::Metrics`, making the contract invisible and teaching implementors that `metrics` is a fire-and-forget side-effecting function. Any agent reading this file as a template will write adapters that omit the return type on non-`()` Metrics types, causing a compile error that is hard to trace back here.

Fix: add the explicit return annotation.

```rust
fn metrics(&self, _state: &MinimalState) -> Self::Metrics {}
```

---

### [MUST FIX] `update()` is never exercised

`update` drives the timer-based state machine — it is the scheduler's primary re-entry point (see ARCHITECTURE.md: "`update` drives timer-based state transitions"). It is declared on the trait and implemented on `MinimalProtocol`, but zero tests call it. At minimum: one test that calls `update` on a fresh state and asserts the return value. For a quiescent-after-first-poll protocol, `update` returning `None` is the expected contract.

Suggested test:

```rust
#[test]
fn protocol_update_returns_none_when_quiescent() {
    let protocol = MinimalProtocol;
    let (mut state, _) = protocol.init(());
    // Drain the single transmission so the protocol is quiescent.
    let _ = protocol.poll_transmit(&mut state, 0);
    let next_wake = protocol.update(&mut state, 1);
    assert!(
        next_wake.is_none(),
        "update() should return None for a quiescent protocol with no pending timers"
    );
}
```

---

### [MUST FIX] `on_receive` test does not assert on the return value

The contract for `on_receive` is `-> Option<SimTime>` (the next wake time). The test only checks no-panic:

```rust
let _ = protocol.on_receive(&mut state, frame, 100);
```

Discarding the return value with `let _` makes the test blind to the return value. For `MinimalProtocol`, the expected return is `None` (no timer scheduled). Add an assertion:

```rust
let wake = protocol.on_receive(&mut state, frame, 100);
assert!(
    wake.is_none(),
    "on_receive() should return None when no timer is scheduled"
);
```

---

### [SHOULD FIX] `init` test is tautological

`protocol_init_provides_wake_time` asserts `wake_time.is_some()`, which is trivially true because `MinimalProtocol::init` hard-codes `Some(0)`. This test does not exercise the contract — it re-asserts what the stub happens to return.

The useful contract here is the converse: a protocol that does NOT plan to transmit at startup should return `None`. Without a second stub (a `SilentProtocol` that returns `None` from `init`), neither branch of the contract is meaningfully tested. Consider either:

- Adding a `SilentProtocol` stub that returns `(state, None)` from `init` and asserting `wake_time.is_none()` in a paired test, or
- Renaming the current test to `protocol_with_pending_tx_init_provides_wake_time` to make the scope explicit and adding the `None` case.

---

### [INFORMATIONAL] Codecov gate is failing

The Codecov bot reports coverage at 99.49%, below the 100% configured target. This is a pre-existing miss (1 uncovered line on `main` before this PR), but the gate is broken and this PR cannot merge cleanly while the Codecov status check fails. The PR description should acknowledge this, or the existing uncovered line should be addressed before or alongside this PR.

---

### [INFORMATIONAL] No linked issue

The PR description does not reference a parent issue. Related open issues found in the repo: the Kani verification epic (#30) and the protocol trait Kani proofs issue (#37). If this PR is groundwork for #37 or similar, linking helps agents working in parallel avoid duplicating the stub implementor.

---

### [MINOR] PR description uses "idempotence" incorrectly

The description says "poll_transmit idempotence" but the test actually checks *draining* (first call returns `Some`, second returns `None`). Idempotence means repeated calls return the same value — the opposite of what is being tested. The test name `protocol_poll_transmit_once_then_none` is correct; the description should be updated to "poll_transmit drain contract" or similar.

---

PR #28, opened 2026-04-04
refactor: single-pass node delivery + remove dead bandwidth field

DominicBurkart:
## Summary

- **Single-pass delivery in `deliver_completed_to_nodes()`** (`src/scheduler.rs`): the non-collision path previously iterated over `self.nodes` three times — once for `on_receive()` + wake collection, once to schedule wakes, and once more to collect indices for `handle_poll_transmit()`. All three passes are now consolidated into a single loop that calls `on_receive()`, immediately schedules any returned wake time, and calls `handle_poll_transmit()` within the same iteration. The two intermediate `Vec` allocations (`wakes` and `tx_node_idxs`) are eliminated.

- **Remove dead `bandwidth` field from `ActiveTransmission`** (`src/channel.rs`): the field was stored under `#[allow(dead_code)]` but never read after construction. The field and its lint suppression are removed, and the construction site in `begin_transmission()` no longer copies `tx.bandwidth` into the internal struct. The `bandwidth` field on the public `Transmission` type is unchanged.

## Test plan

- [ ] `cargo check` passes (verified before push)
- [ ] `cargo test` passes with no regressions across channel and scheduler unit/property tests
- [ ] Confirm `total_rx`, `total_tx`, collision, capture, and airtime metrics are identical to pre-refactor behaviour across existing test cases

https://claude.ai/code/session_01PBsbV6qt74AVdmi4nfo9nD

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -3,6 +3,82 @@ use crate::types::{ChannelEvent, NodeId, RxMetadata, Transmission};
 
 pub type CompletedTx = (NodeId, bool, bool, RxMetadata);
 
+/// Default LoRa path loss in dB (free-space + typical indoor attenuation baseline).
+pub const LORA_PATH_LOSS_DB: f32 = 100.0;
+
+/// Default LoRa noise floor in dBm (LoRa sensitivity at SF7/125 kHz).
+pub const LORA_NOISE_FLOOR_DBM: f32 = -117.0;
+
+/// Default co-channel rejection threshold in dB (LoRa capture effect threshold).
+pub const LORA_CO_CHANNEL_REJECTION_DB: f32 = 6.0;
+
+/// Configuration for physical-layer channel parameters.
+///
+/// Construct with [`ChannelConfig::lora_defaults()`] for LoRa or supply
+/// custom values for other protocols.
+///
+/// # Examples
+///
+/// ```
+/// use theatron::channel::ChannelConfig;
+///
+/// // LoRa defaults
+/// let cfg = ChannelConfig::lora_defaults();
+///
+/// // Custom 802.15.4 / Zigbee-like parameters
+/// let cfg = ChannelConfig {
+///     path_loss_db: 80.0,
+///     noise_floor_dbm: -100.0,
+///     co_channel_rejection_db: 3.0,
+/// };
+/// ```
+#[derive(Debug, Clone, PartialEq)]
+pub struct ChannelConfig {
+    /// Path loss applied to every transmission (dB). Higher values mean
+    /// the received signal is weaker.
+    pub path_loss_db: f32,
+
+    /// Noise floor of the receiver (dBm). Used to compute SNR.
+    pub noise_floor_dbm: f32,
+
+    /// Minimum power difference (dB) required for the stronger signal to
+    /// survive a co-channel collision via the capture effect.
+    pub co_channel_rejection_db: f32,
+}
+
+impl ChannelConfig {
+    /// Return the LoRa default parameters used by [`Channel::new()`].
+    ///
+    /// | Parameter               | Value    |
+    /// |-------------------------|-----------|
+    /// | `path_loss_db`          | 100 dB   |
+    /// | `noise_floor_dbm`       | -117 dBm |
+    /// | `co_channel_rejection_db` | 6 dB   |
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use theatron::channel::ChannelConfig;
+    /// let cfg = ChannelConfig::lora_defaults();
+    /// assert_eq!(cfg.path_loss_db, 100.0);
+    /// assert_eq!(cfg.noise_floor_dbm, -117.0);
+    /// assert_eq!(cfg.co_channel_rejection_db, 6.0);
+    /// ```
+    pub fn lora_defaults() -> Self {
+        Self {
+            path_loss_db: LORA_PATH_LOSS_DB,
+            noise_floor_dbm: LORA_NOISE_FLOOR_DBM,
+            co_channel_rejection_db: LORA_CO_CHANNEL_REJECTION_DB,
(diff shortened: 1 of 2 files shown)


DominicBurkart (commented):
## Code Review — `refactor: single-pass node delivery + remove dead bandwidth field`

Audience: coding agents. PR is a draft; review covers correctness, semantic safety, test coverage, and architectural alignment.

---

### Summary

Both changes are correct and safe. The single-pass refactor eliminates two intermediate `Vec` allocations and three list traversals without altering observable simulation behaviour. The dead `bandwidth` field removal is a clean hygiene fix. Coverage is confirmed at 100% by Codecov. No secrets, PII, or architectural drift detected.

---

### Finding 1 — IMPORTANT: Execution-order semantic shift is safe TODAY but fragile under future `NodeHandle` evolution

**File:** `src/scheduler.rs`, the consolidated loop in `deliver_completed_to_nodes`.

**What changed:** In the old code all `on_receive()` calls completed before any `handle_poll_transmit()` call. In the new code, `handle_poll_transmit(i, time)` is called for node `i` before `on_receive()` is called for node `i+1`.

This is currently safe because:
- `on_receive` receives only an `RxMetadata` value (no `Channel` access).
- `handle_poll_transmit` adds to `channel.active` but does not affect `channel.completed`, which has already been drained.
- New `TxComplete` events enqueued during this loop cannot fire until after `run()` returns control to its event-pop loop.

However, the invariant relied upon — "a node's `poll_transmit` cannot produce any observable effect on the `on_receive` call of a later node within the same delivery pass" — is not enforced by the type system or documented in `NodeHandle`'s contract. A future implementor who, say, reads shared state in `on_receive` that is mutated by `poll_transmit` (e.g. via an `Arc<Mutex<...>>` in a test harness or the future `lorawan-device` adapter) could encounter non-deterministic behaviour depending on their index in `self.nodes`.

**Recommendation:** Add a comment at the loop head in `deliver_completed_to_nodes` documenting this ordering assumption, e.g.:

```rust
// Single-pass delivery: on_receive(i) → schedule_wake(i) → handle_poll_transmit(i)
// before moving to node i+1. This is safe because on_receive receives only
// a cloned RxMetadata value and does not observe channel state mutated by
// handle_poll_transmit. Protocol implementations that share mutable state
// across on_receive and poll_transmit must not depend on all on_receive calls
// completing before any poll_transmit call.
```

---

### Finding 2 — MINOR: `handle_poll_transmit` is called even when `on_receive` returns `None`; this matches old behaviour but consider whether it should be conditional

In both old and new code, `handle_poll_transmit(i, time)` is called unconditionally for every non-sender node on every frame delivery, regardless of whether `on_receive` returned a wake time or indicated any state change. This means every delivery event polls every receiver for a pending transmission.

This is intentional based on `ReplyNode` semantics (a node sets a flag in `on_receive` and then `poll_transmit` picks it up immediately). The refactor preserves this. However, it means O(N) `poll_transmit` calls per completed frame even for idle nodes. This is not a regression introduced by this PR — it pre-existed — but it is worth flagging now since the refactor touches this code path and the project is still pre-1.0. As node counts grow (the architecture roadmap targets scalability evaluation), this becomes a hotspot. A future optimisation would be an explicit "dirty" flag or a separate ready-queue for nodes that have pending TX.

No action required in this PR; noting for backlog.

---

### Finding 3 — LOW: `deliver_to` on `Channel` is now unreachable from `Scheduler`

`channel.rs` exposes a `deliver_to(&self, time: SimTime) -> Vec<RxMetadata>` method (used in many `channel.rs` unit tests) but `Scheduler` exclusively uses `drain_completed()`. After this PR, `deliver_to` is tested only at the `Channel` unit-test level and is not exercised by any scheduler-level integration path. This is not a regression from this PR — it was already the case — but the removal of the `bandwidth` field touched the same struct, making this a natural moment to flag.

If `deliver_to` is intended as part of `Channel`'s public API for external users building custom schedulers, it should be `pub` (it already is) and documented. If it is only used internally for tests, it could be gated with `#[cfg(test)]` to signal that intent. No action required in this PR.

---

### Verdict

No correctness bugs. The refactor is semantically equivalent under the current `NodeHandle` contract. The `bandwidth` dead-field removal is correct — `Transmission.bandwidth` (public) is unchanged; only `ActiveTransmission.bandwidth` (private, never read) is removed. 100% test coverage confirmed. This PR is ready to merge once the draft state is lifted and the test-plan checklist is verified by a human or CI run.

One optional polish item: the comment in Finding 1 would make the ordering invariant explicit for future agents implementing `NodeHandle`.

---

DominicBurkart:
Addressed Finding 1 (IMPORTANT) from @DominicBurkart's review.

**What was added:** A 5-line comment block directly above the delivery loop in `deliver_completed_to_nodes` (`src/scheduler.rs`) documenting the single-pass ordering invariant:

```rust
// Single-pass delivery: on_receive(i) → schedule_wake(i) → handle_poll_transmit(i)
// before moving to node i+1. This is safe because on_receive receives only
// a cloned RxMetadata value and does not observe channel state mutated by
// handle_poll_transmit. Protocol implementations that share mutable state
// across on_receive and poll_transmit must not depend on all on_receive calls
// completing before any poll_transmit call.
```

No logic was changed — only the comment was added. Findings 2 and 3 were marked "no action required in this PR" and were not acted upon.

https://claude.ai/code/session_019T41CUQKAtDfzRrZkWQTFW

---

DominicBurkart:
All three subagents completed. The branch had a real merge conflict — it was based on the old flat-field `Channel` struct before the `ChannelConfig` API refactor landed via PR #21; this session rebased both intended changes (single-pass `deliver_completed_to_nodes` loop eliminating the `receiver_results` Vec, and `bandwidth` field removal) onto main's current API and pushed commit `50476d7`. The PR has been promoted from draft to open for human review.

---

PR #41, opened 2026-04-05
test: add Protocol trait and scheduler timer-contract tests

DominicBurkart:
## Problem

The `Protocol` trait in `src/traits.rs` is the central abstraction of theatron — every wireless-protocol implementation must implement it — but it had **zero test coverage**. No existing test instantiates `Protocol` directly, and the scheduler's key timer-delivery invariant was never verified.

## Timer-delivery invariant

> If `Protocol::update` (or `NodeHandle::update`) returns `Some(t)`, the scheduler **must** call `update` again at exactly time `t` — no earlier, no later.

This invariant is the foundation of any protocol that uses deferred wake-ups (RX windows, back-off timers, duty-cycle enforcement, etc.).  A regression in `Scheduler::run`'s wake-scheduling path would silently break all such protocols.

### `ProtocolNode<P: Protocol>` bridge

A generic wrapper that pairs a `Protocol` implementation with its mutable `State` and implements `NodeHandle`, allowing any `Protocol` to be registered with `Scheduler`.  This is the missing glue layer between the stateless `Protocol` trait and the stateful `NodeHandle` interface.

### `TwoPhaseProtocol`

A minimal `Protocol` that:
1. At t = 0 transmits one beacon frame and returns `Some(1_000_000)` from `update` to schedule an RX window.
2. At t = 1_000_000 records the exact time it was called.

### Tests

| Test | What it checks |
|---|---|
| `scheduler_delivers_wake_at_exact_scheduled_time` | Timer contract: `update` is called at exactly t = 1_000_000; `current_time() >= 1_000_000`; exactly 1 TX recorded in metrics |
| `protocol_node_no_op_does_not_panic` | A `Protocol` that never transmits/wakes runs cleanly with 0 TX |
| `protocol_init_wake_at_zero_fires_update` | `init` returning `Some(0)` causes `update` to fire before simulation ends |
| `main_binary_is_noop_and_should_be_removed` | Documentation-only: flags that `src/main.rs` and its `main_does_not_panic` test are trivial stubs |

## Recommended follow-up

`src/main.rs` currently contains only `println!("theatron")` and a `main_does_not_panic` test that adds no value.  The `[[bin]]` target and that file should be removed (along with the corresponding entry in `Cargo.toml`) once the project no longer needs a binary target.  The `main_binary_is_noop_and_should_be_removed` test in this file documents that intent without causing a build failure.

### DIFF
--- a/tests/protocol_timer_contract.rs
+++ b/tests/protocol_timer_contract.rs
@@ -1,0 +1,382 @@
+//! Tests for the `Protocol` trait contract and the scheduler's timer-delivery invariant.
+//!
+//! # What is the timer contract?
+//!
+//! When `Protocol::update` (or `NodeHandle::update`) returns `Some(t)`, the scheduler
+//! is required to call `update` again at *exactly* time `t` — no earlier, no later.
+//! These tests verify that invariant holds end-to-end through `Scheduler::run`.
+//!
+//! # Why does this matter?
+//!
+//! The `Protocol` trait is the central abstraction in theatron, yet it had zero test
+//! coverage before this file.  Any regression in the scheduler's wake-scheduling path
+//! could silently break all protocol implementations.
+
+use theatron::scheduler::{NodeHandle, Scheduler};
+use theatron::time::SimTime;
+use theatron::traits::Protocol;
+use theatron::types::{NodeId, RxMetadata, Transmission};
+
+// ---------------------------------------------------------------------------
+// TwoPhaseProtocol — a minimal Protocol implementation used by the tests
+// ---------------------------------------------------------------------------
+
+/// A two-phase protocol that:
+///
+/// 1. At t = 0 transmits one beacon frame and asks to be woken at t = 1_000_000.
+/// 2. At t = 1_000_000 records the exact wake time so the test can assert it.
+///
+/// This is intentionally the smallest possible `Protocol` that exercises both
+/// the transmit path and the deferred-wake (timer) path.
+struct TwoPhaseProtocol;
+
+struct TwoPhaseState {
+    /// Set to `true` once `update` has been called at the RX-window time.
+    rx_window_fired: bool,
+    /// The `SimTime` value passed to the second `update` call (the RX window).
+    rx_window_time: Option<SimTime>,
+    /// Number of frames queued for transmission by `poll_transmit`.
+    transmit_count: u32,
+}
+
+impl Protocol for TwoPhaseProtocol {
+    type Config = ();
+    type State = TwoPhaseState;
+    type Metrics = u32; // returns the transmit count
+
+    fn init(&self, _config: ()) -> (TwoPhaseState, Option<SimTime>) {
+        let state = TwoPhaseState {
+            rx_window_fired: false,
+            rx_window_time: None,
+            transmit_count: 0,
+        };
+        // Wake immediately at t = 0 so `update` → `poll_transmit` fires.
+        (state, Some(0))
+    }
+
+    fn on_receive(
+        &self,
+        _state: &mut TwoPhaseState,
+        _frame: RxMetadata,
+        _time: SimTime,
+    ) -> Option<SimTime> {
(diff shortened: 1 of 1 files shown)


DominicBurkart (commented):
## Code Review — `tests/protocol_timer_contract.rs`

All 6 CI checks pass (Autoformat, Cargo Deny, Security Audit, Check, Test, Code Coverage). No previous review exists on this PR. The branch is marked **Draft**.

---

### Summary

This PR closes a real gap: the `Protocol` trait had zero integration test coverage, and the scheduler's timer-delivery invariant was entirely untested. The intent, the PR description, and the overall structure are all good. The issues below are worth addressing before the PR is un-drafted.

---

### Issues

#### 1. `TwoPhaseProtocol` and `ProtocolNode` are defined but never used in any test

`TwoPhaseProtocol` (lines 32–110) and the first `ProtocolNode::new` call (line 145) in `scheduler_delivers_wake_at_exact_scheduled_time` are dead code. The test pivots mid-function to `ObservingProtocol` (with the comment "Re-implement with shared state") and the original `node`/`initial_wake` variables from line 145 are never referenced again. This will produce compiler warnings and, more importantly, leaves `TwoPhaseProtocol` completely untested despite being the most carefully documented type in the file.

**Options:**
- Remove `TwoPhaseProtocol` and the dead `ProtocolNode::new` call entirely, or
- Use `TwoPhaseProtocol` directly in a separate test that inspects its `metrics()` return value (transmit count), and drop `ObservingProtocol` in favour of that approach.

#### 2. The large comment block in `scheduler_delivers_wake_at_exact_scheduled_time` documents rejected approaches inline

Lines 150–173 are a multi-paragraph comment that walks through three abandoned implementation strategies (`unsafe` pointer aliasing, `Box<dyn NodeHandle>` downcasting, `Rc<RefCell<…>>`), then concludes by choosing the third. This is useful design reasoning for a PR description but is noise in shipped test code — it will confuse future readers who expect the comment to describe what the code *does*, not what was tried and discarded.

**Suggestion:** Collapse to a single-sentence rationale: `// Rc<RefCell<…>> allows post-run observation of protocol state without unsafe.`

#### 3. `main_binary_is_noop_and_should_be_removed` is a test that unconditionally passes and asserts nothing

A test whose body is only comments and whose only behaviour is `pass` does not belong in a test suite — it cannot catch regressions and it inflates test-count metrics. The stated intent (document that `src/main.rs` should be removed) is better served by:
- A `// TODO:` comment in `src/main.rs` itself, or
- A tracking issue linked from the PR description (which already describes the problem clearly).

#### 4. `ObservingProtocol` uses `Rc<RefCell<…>>` for `State` but `Protocol::State = ()` — the type signature is misleading

The `Protocol` trait's `State` associated type is the idiomatic place to carry mutable per-node state. Using `()` for `State` while routing all real state through a closure-captured `Rc<RefCell<…>>` works but contradicts the abstraction the trait is trying to enforce. A reader learning from these tests will get the wrong model of how `Protocol` is intended to be implemented. Consider using `Rc<RefCell<ObservingState>>` as the `State` type directly, with `ProtocolNode` storing it, so the pattern matches the trait's design intent.

#### 5. Minor: `_node` and `_initial_wake` (the dead bindings at line 145) should trigger a Clippy warning

If they do not trigger a warning today it is only because `TwoPhaseProtocol` happens to also be used implicitly. After fixing issue #1 the compiler should flag these; confirm with `cargo clippy --tests` before marking ready for review.

---

### Positive notes

- The `ProtocolNode` bridge struct is the right abstraction and should be promoted to the library proper (e.g. `src/protocol_node.rs`) so that users don't have to rewrite it in every integration test.
- `protocol_node_no_op_does_not_panic` and `protocol_init_wake_at_zero_fires_update` are clean, focused, and correctly structured.
- The module-level doc comment accurately describes the timer contract and why it matters — keep it.
- All CI green, single-commit, clean diff.


---

PR #56, opened 2026-04-17
arch(theatron): reduce complexity and improve alignment

DominicBurkart:
## Issues addressed

- **Duplicate node-list iteration in `deliver_completed_to_nodes`** — the method looped over all nodes twice with the identical `node_id() != sender` guard: once for `on_receive` + wake collection, then again to build `tx_node_idxs`. Merged into a single pass that collects both in one traversal.
- **Dead `sender` field on `TxComplete` event** — `EventKind::TxComplete { sender }` carried a `NodeId` that was always bound to `_` at every match site and never read. Replaced with a unit-like variant `TxComplete` to eliminate the field and the dead construction of it in both `handle_poll_transmit` and the `InterferencePoll` arm.
- **Unnecessary `find_node_idx` helper** — a private one-call wrapper around `Iterator::position` that added a layer of indirection without clarity benefit. Inlined the `position` call directly into the `Wake` match arm where it was used.

### DIFF
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -53,7 +53,8 @@ pub trait NodeHandle {
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub enum EventKind {
     Wake { node_id: NodeId },
-    TxComplete { sender: NodeId },
+    /// Fired when a transmission window closes; the channel resolves overlap at this point.
+    TxComplete,
     InterferencePoll { interferer_idx: usize },
 }
 
@@ -168,10 +169,6 @@ impl Scheduler {
         self.events.push(ScheduledEvent { time, seq, kind });
     }
 
-    fn find_node_idx(&self, id: NodeId) -> Option<usize> {
-        self.nodes.iter().position(|n| n.node_id() == id)
-    }
-
     fn handle_poll_transmit(&mut self, node_idx: usize, time: SimTime) {
         if let Some(tx) = self.nodes[node_idx].poll_transmit(time) {
             let sender = self.nodes[node_idx].node_id();
@@ -183,7 +180,7 @@ impl Scheduler {
             self.metrics.record_tx(sender);
             self.metrics.record_airtime(duration);
             let complete_time = time + duration;
-            self.schedule(complete_time, EventKind::TxComplete { sender });
+            self.schedule(complete_time, EventKind::TxComplete);
         }
     }
 
@@ -196,7 +193,10 @@ impl Scheduler {
                 if captured {
                     self.metrics.record_capture();
                 }
+                // Single pass: deliver frame, collect wake requests, and queue
+                // poll_transmit checks — avoids iterating the node list twice.
                 let mut wakes = Vec::new();
+                let mut tx_node_idxs = Vec::new();
                 for i in 0..self.nodes.len() {
                     if self.nodes[i].node_id() != sender {
                         let next = self.nodes[i].on_receive(frame.clone(), time);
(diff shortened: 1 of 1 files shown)


src/scheduler.rs
```diff
@@ -53,7 +53,8 @@ pub trait NodeHandle {
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub enum EventKind {
     Wake { node_id: NodeId },
-    TxComplete { sender: NodeId },
+    /// Fired when a transmission window closes; the channel resolves overlap at this point.
```
DominicBurkart, line 56: `EventKind` is a `pub enum`, so removing the `sender` field from `TxComplete` is a breaking change to a public API surface — not flagged in the PR title/body and no CHANGELOG entry. ARCHITECTURE.md doesn't position `EventKind` as an internal-only type, and the doctest on the enum exercises `Wake` publicly, implying external coding agents may pattern-match on all variants. Consider either (a) noting the breaking change in the PR description and bumping the semver/version, or (b) making `EventKind` `#[non_exhaustive]` going forward so similar trims are non-breaking. The change itself is correct: the field was genuinely dead at every match site.

---

src/scheduler.rs
```diff
@@ -196,25 +193,23 @@ impl Scheduler {
...
                         self.metrics.record_rx(self.nodes[i].node_id());
                         if let Some(t) = next {
                             wakes.push((self.nodes[i].node_id(), t));
                         }
+                        tx_node_idxs.push(i);
                     }
                 }
                 for (node_id, t) in wakes {
                     self.schedule(t, EventKind::Wake { node_id });
                 }
```
DominicBurkart, line 212: The merged single-pass loop is a genuine complexity reduction (one O(N) walk instead of two), and the guard condition is identical so semantics are preserved. Minor nit for future cleanup: `tx_node_idxs` is now exactly equivalent to the set of `i` for which a wake *may or may not* have been pushed — i.e. the same iteration. You could drop the intermediate `Vec<usize>` by calling `self.handle_poll_transmit(i, time)` directly in the same loop, but borrow-checker constraints (`&mut self.nodes[i]` while iterating, then `&mut self` for `handle_poll_transmit`) likely motivated the buffer. Worth a comment acknowledging the borrow-checker reason rather than the perf reason currently implied by "avoids iterating the node list twice".

---

src/scheduler.rs
```diff
@@ -253,15 +248,15 @@ impl Scheduler {
 
             match event.kind {
                 EventKind::Wake { node_id } => {
-                    if let Some(idx) = self.find_node_idx(node_id) {
+                    if let Some(idx) = self.nodes.iter().position(|n| n.node_id() == node_id) {
```
DominicBurkart, line 251: Inlining `find_node_idx` is fine at the single call site, but note ARCHITECTURE.md anticipates growth (multi-protocol comparison in Phase 2, parameter sweeps in Phase 4) where node-index lookup will likely recur. If a second caller appears, re-extracting is cheap — but at that point consider a `HashMap<NodeId, usize>` rather than reintroducing linear `position` scans, since node counts are expected to grow meaningfully per the scalability evaluation dimension. Not blocking for this PR.

---

DominicBurkart (commented):
Overall this is a legitimate complexity reduction, not complexity motion: the three changes eliminate dead data, a redundant traversal, and a thin helper. No tests were deleted; the existing suite (14+ unit tests, 1 proptest) still covers all affected paths. Alignment with ARCHITECTURE.md is good — scheduler stays an event-driven priority queue as specified, and nothing touches the `Protocol`/`TrafficModel` contracts.

Main concern: `EventKind` is `pub`, so trimming `TxComplete`'s field is technically a breaking change and should be called out in the PR description (or the enum made `#[non_exhaustive]`). Minor doc nit on the merged loop's comment and a forward-looking note on node-index lookup if it recurs. Nothing blocking.

---

PR #59, opened 2026-04-20
test: integration + proptest coverage for AlohaNode example

DominicBurkart:
Automated test improvement PR — please review.

## Component chosen
`examples/aloha/aloha_node.rs` — `AlohaNode`, `AlohaReceiver`, and the
`PeriodicTraffic` traffic model.

## Current state
- `PeriodicTraffic` has two focused unit tests in the example file.
- `AlohaNode` has two shallow unit tests; `AlohaReceiver` has one.
  None drive the types through a real `Scheduler`.
- `tests/aloha.rs` explicitly documents that these types are **not**
  exercised end-to-end and uses hand-written stand-ins instead. The
  file's own doc-comment recommends the `#[path = ...]` approach used
  by this PR as the way to close the gap.

## Strategy
- **Realistic integration** (scheduler + channel + example types) for
  the TX-count, payload-integrity, and SF-orthogonality invariants.
- **Proptest** for the two core `PeriodicTraffic` contracts (exact
  count and interval gating) that every protocol using the model
  relies on.
- No doctests — these types live in the example crate, not the
  public theatron API.

## Invariants validated
1. A single-payload `AlohaNode` drives exactly one TX end-to-end and
   halts the scheduler (no infinite wake loop) when traffic is
   exhausted.
2. The TX payload is delivered verbatim to a passive `AlohaReceiver`.
3. Two simultaneous `AlohaNode`s on orthogonal SFs both deliver with
   zero collisions.
4. `PeriodicTraffic::next_payload` emits at most `count` payloads
   total and gates on the configured interval.

## Out of scope
While writing these tests I observed a latent interaction between
`AlohaNode::update`'s "probe" path and `PeriodicTraffic`'s stateful
`next_payload`: probing a future time consumes a slot that is never
stored, so multi-payload `count > 1` scenarios silently drop
payloads. That is a product bug, not a test posture issue, so these
tests pin down the `count = 1` regime and the `PeriodicTraffic`
model in isolation. Filing the bug separately is recommended.

## Test plan
- [x] `cargo test --test aloha_node_integration` — 10/10 pass locally
  (5 new + 5 from the imported example module).
- [x] `cargo fmt --check` — new file is clean.
- [ ] CI pipeline green.

https://claude.ai/code/session_01KcPL3wPfBJhiKAzBB9Lvd5


DominicBurkart (commented):
## Review — `tests/aloha_node_integration.rs`

Overall: the structure is sound and the gap in `tests/aloha.rs` is correctly identified. Two items need resolution before merge (payload verbatim assertion is missing; misleading inline comment in production code), one confirmed bug needs a tracking issue, and several notes below are informational.

---

### Confirmed: the probe/PeriodicTraffic bug is real

The PR body describes it correctly. Walk-through for `count = 2`:

1. `AlohaNode::update(0)` → `try_generate_tx(0)`.
2. `next_payload(0)` succeeds: `remaining 2 → 1`, `next_time = interval_us`, payload stored in `pending_tx`.
3. Scheduler calls `poll_transmit` — TX fires, `pending_tx` cleared.
4. Scheduler calls `update` again at a time `< interval_us`.
5. `pending_tx` is `None`; `next_payload(t)` returns `None` (too early).
6. **Probe fires**: `next_payload(interval_us)` — condition `interval_us >= next_time` is true, so `remaining 1 → 0`, `next_time = 2 * interval_us`. **The slot is consumed and the payload is discarded.**
7. Probe returned `Some(...)` so `try_generate_tx` returns `Some(future)` — node stays scheduled.
8. At the next wake, `next_payload(interval_us)` → `remaining == 0` → `None`. Probe at `2 * interval_us` → `None`. Node returns `None` and stops. Second payload is silently dropped and never transmitted.

The inline comment at the probe call site in `examples/aloha/aloha_node.rs` reads:

> `next_payload` consumed the future slot, but `PeriodicTraffic` re-checks `time >= next_time`, so calling it early is safe.

This comment is **incorrect**: the slot is consumed (`remaining` is decremented). It needs to be corrected or removed, and a bug issue should be filed before this PR merges so the follow-up is tracked. The tests correctly pin `count = 1` to avoid this, but the misleading comment will confuse future readers.

**Action required**: correct or remove the comment in `try_generate_tx`; open (and reference) a tracking issue for the probe bug.

---

### Required — test (2) does not assert payload bytes

The module-level doc says invariant (2) is:

> The TX payload is delivered verbatim to a passive `AlohaReceiver`.

But `single_shot_aloha_delivers_payload_to_receiver` only checks `node_rx_count(NodeId(99)) == 1`. It does not inspect the received bytes. `AlohaReceiver::received` is `pub Vec<RxMetadata>`, so if the scheduler exposes nodes after `run()` the assertion is straightforward:

```rust
let rxer: &AlohaReceiver = /* downcast from scheduler */;
assert_eq!(rxer.received[0].payload, payload);
```

If the scheduler does not expose nodes post-run, the test should document that limitation explicitly and the invariant description should be changed from "verbatim" to "delivery confirmed by rx count only." Right now the doc and the assertion are inconsistent.

**Action required**: either add the payload-bytes assertion or narrow the stated invariant to match what is actually checked.

---

### Informational — `prop_assume!` is unreachable

In `periodic_traffic_interval_gates_next_payload`:

```rust
let too_early = (interval_us * early_offset) / 100;
prop_assume!(too_early < interval_us);
```

With `early_offset in 1u64..=99`, the maximum value of `too_early` is `interval_us * 99 / 100`, which by integer division is always `< interval_us`. The guard never fires and shrinks no cases. It is harmless but dead — remove it or add a comment explaining the arithmetic guarantee.

---

### Informational — proptest covers the model, not the integration path

`periodic_traffic_yields_exactly_count_payloads` drives `PeriodicTraffic` directly at perfectly aligned intervals. It correctly tests the model in isolation and intentionally avoids the probe path. This is the right scope given the count=1 pinning strategy. A future proptest that drives `AlohaNode::update` directly with `count=1` and asserts no double-emit would close the remaining gap without depending on the probe bug.

---

### Informational — `#[path = ...]` approach is sound

`aloha_node.rs` uses only `theatron::...` imports, which resolve correctly when compiled as a submodule of the integration test crate (which has `theatron` as a dev-dependency). No `extern crate` or relative module paths inside the file. The pattern is safe here and matches what `tests/aloha.rs` recommends.

---

### Informational — halt assertion is a meaningful pin

The `sched.current_time() < end_time` assertion in test (1) validates that the `None`-return path in `try_generate_tx` actually stops the scheduler from looping indefinitely. Keep it.

---

### Summary of actions before merge

1. **Correct the misleading comment** in `examples/aloha/aloha_node.rs` at the probe call site in `try_generate_tx`. Open a bug issue for the probe/slot-consumption interaction and reference it in the comment.
2. **Fix or scope down the "verbatim" claim** in test (2): either assert `received[0].payload == payload` or change the invariant description to match what is actually checked.

https://claude.ai/code/session_01KxYQiFucgGCsPzQ2ALJKLD

---

DominicBurkart:
**CRON status update** — CI is fully green (6/6 checks); a coverage fix was pushed this cycle adding a `count=2` test to cover the `Some(future)` branch in `try_generate_tx`. However, the code review **confirmed the documented probe/PeriodicTraffic bug is real**: `next_payload` is unconditionally stateful, the probe silently consumes the second payload slot, and the inline comment "calling it early is safe" is incorrect and should be removed; additionally, test 2's claim of "payload delivered verbatim" is unsubstantiated (no byte-level assertion). **Not ready for human review** — the confirmed bug must be addressed and test 2 strengthened. [Session](https://claude.ai/code/session_01KxYQiFucgGCsPzQ2ALJKLD)

---

DominicBurkart (commented):
## Review — `tests/aloha_node_integration.rs` (commit 528fc5a)

The second commit adds test (5) (`aloha_node_defers_wake_when_payload_not_yet_ready`), which covers the previously-missing `Some(future)` branch. That is a genuine improvement. One required item from the previous review remains open, and one new concern about test (5) itself.

---

### Required — test (2) still does not assert payload bytes

`single_shot_aloha_delivers_payload_to_receiver` checks only `node_rx_count(NodeId(99)) == 1`. The module-level doc calls invariant (2) "delivered **verbatim**", which implies byte-level equality. The two are inconsistent. Either:

- Assert `received[0].payload == payload` (if `AlohaReceiver::received` is accessible after `sched.run()`), or
- Change the doc to "delivery confirmed by rx count only" and drop the word "verbatim".

This was flagged in the previous review and has not changed.

---

### New concern — test (5) exercises the probe-bug path without flagging it

`aloha_node_defers_wake_when_payload_not_yet_ready` uses `count = 2`. The inline comment correctly traces the probe call consuming `remaining 1 → 0`. After `update(500_000)` returns `Some(1_500_000)`, there are zero remaining payloads — the second slot was silently consumed by the probe. A subsequent `update(1_500_000)` would find nothing and return `None`, so the scheduler would stop and the second payload would be dropped.

The test passes (the `Some(1_500_000)` assertion is correct), but it does not pin down the probe's side-effect. A reader following the test expects a second TX at `t = 1_500_000`; the code actually drops it. Adding a comment such as:

```rust
// NOTE: the probe call consumed the second slot (the known probe/PeriodicTraffic
// bug). A follow-up update at t=1_500_000 would return None, not Some.
// This test only validates the branch return value, not downstream behavior.
```

…prevents the test from becoming a false-confidence signal, and keeps the out-of-scope note in sync with what the test code actually does.

---

### Informational — `prop_assume!` is still unreachable

`too_early = (interval_us * early_offset) / 100` with `early_offset in 1..=99` is always `< interval_us` by integer division. The guard fires on no case and shrinks nothing. Harmless, but dead — remove it or add an arithmetic-guarantee comment.

---

### Summary

| # | Status | Action |
|---|--------|--------|
| Test (2) "verbatim" mismatch | **Still open** | Assert bytes or narrow the doc |
| Test (5) probe-bug side-effect undocumented | **New** | Add clarifying comment |
| `prop_assume!` unreachable | Informational | Clean up |

https://claude.ai/code/session_01RTLA2PX1wUvM9oj8wrAteb

---

DominicBurkart:
**Automated cron status.** Review posted on new head `528fc5a` — 3 issues: (1) verbatim-delivery invariant claims payload correctness but only asserts `node_rx_count == 1` (payload bytes never compared); (2) `count = 2` probe test drops the second slot silently without a follow-up `update` assertion or clarifying comment; (3) the `prop_assume!(too_early < interval_us)` guard is dead code given the strategy bounds. Address-comments agent stalled mid-run this iteration — no commit pushed. CI remains green on the current head; re-run the cleanup iteration to land the 3 fixes — **not yet ready for human review**. [session](https://claude.ai/code/session_01RTLA2PX1wUvM9oj8wrAteb)

---

DominicBurkart:
**Automated cron status** ([session](https://claude.ai/code/session_01Nokmxow6PWSA7TRChp5o8u))

Review subagent skipped (prior review on `528fc5a` is current). Address-comments subagent pushed `9a8581d` addressing all 3 actionable items: invariant (2) doc narrowed to rx-count confirmation, test (5) gained a NOTE about destructive probe consumption, and the unreachable `prop_assume!` was replaced with an arithmetic-guarantee comment; local `cargo test --all` 35/35 + `aloha_node_integration` 11/11. CI was green on the prior commit and the changes are test-only — once the new commit goes green this is **READY**; leaving draft pending CI recheck.

---

PR #61, opened 2026-04-21
test: lock in scheduler ordering + end_time invariants (plus dead-code debloat)

DominicBurkart:
## Component

`src/scheduler.rs` — the `Scheduler` event loop, specifically the FIFO tiebreaker inside `ScheduledEvent::cmp` and the `event.time > end_time` termination check in `Scheduler::run`.

## Current disposition

Before this PR, `tests/core_coverage.rs::scheduler_determinism` was the only indirect probe of run-to-run reproducibility (metrics-level, a single scenario), and the `end_time` boundary was only exercised with `<= end_time` scenarios. No test asserted:

- simultaneous events fire in insertion order — the `seq` tiebreaker is load-bearing because `std::collections::BinaryHeap` does not document stable ordering;
- a wake at exactly `end_time` fires (`>` vs `>=` off-by-one);
- a wake one tick past `end_time` is dropped;
- `add_interferer(…, first_poll = end_time + 1)` correctly drops the poll;
- broadcast delivery visits receivers in registration order (required for determinism if `self.nodes` ever changes representation);
- self-rescheduling periodic nodes halt exactly at the boundary.

Any of these regressing would silently corrupt the project's determinism promise (called out in ARCHITECTURE.md as foundational) without triggering a test failure.

## Invariants under test

1. **FIFO among simultaneous events**: events scheduled with equal `SimTime` fire in push order (via `seq`).
2. **Run-to-run determinism**: identical inputs produce bitwise-identical event traces.
3. **`end_time` boundary inclusivity**: loop breaks on `>`, so `event.time == end_time` fires; `event.time == end_time + 1` does not.
4. **Interferer termination**: an interferer whose `first_poll > end_time` never has `poll_inject` called.
5. **Broadcast receiver-visit order**: `deliver_completed_to_nodes` calls `on_receive` on registered nodes in order.
6. **Periodic termination**: a node returning `Some(t + period)` from `update` fires at every multiple `<= end_time` and stops.

## Strategy and why

**Strategy**: integration tests in `tests/scheduler_ordering.rs` using `Rc<RefCell<Vec<_>>>` probes attached to custom `NodeHandle` / `InterferenceSource` implementations to capture the exact sequence of scheduler callbacks.

**Why not proptest**: the invariants are deterministic boolean predicates about event ordering at specific timestamps — concrete small cases are more expressive, debuggable, and cheaper than randomized ones. Generating valid interleaved scenarios would also require modeling the scheduler's state machine in the generator, duplicating the thing under test.

**Why not unit tests in `src/scheduler.rs`**: the assertions are about observable behaviour of the public API (`Scheduler::run`, `add_node`, `add_interferer`), so integration scope is right and demonstrates what external users actually depend on. Tests also serve as executable documentation for invariants not spelled out inline in the scheduler source.

## Also: dead-code debloat (commit 54876e6)

- `tests/protocol_timer_contract.rs`: `TwoPhaseProtocol`/`TwoPhaseState` (~85 lines) were defined but the one test that instantiated them immediately shadowed the binding and used an inline `ObservingProtocol` with `Rc<RefCell<_>>` shared state instead — all dead code removed. The same test had ~15 lines of decision-narrative comments ("Simpler approach:", "Actually the cleanest approach:", ...) documenting abandoned designs; replaced with a 4-line rationale for the chosen approach. `main_binary_is_noop_and_should_be_removed` was a no-op test whose body was only a comment about another file — the observation belongs in a PR body.
- `src/scheduler.rs`: drop unused `ChannelConfig` import (the tests module re-imports locally).
- `src/channel.rs::Channel::deliver_to` computed `compute_rssi` twice per element; cached once and passed to `compute_snr`.
- `examples/lorawan_file_transfer/simulated_radio.rs`: move `Bandwidth`/`CodingRate`/`SpreadingFactor` imports into `mod tests` since top-level code only needs `BaseBandModulationParams`.

## Test plan

- [x] `cargo test --all-features` — all suites pass (new `scheduler_ordering` suite: 7 tests)
- [x] `cargo clippy --all-features --all-targets -- -D warnings`
- [x] `cargo fmt --all -- --check`
- [x] pre-commit hooks (fmt, clippy, test, coverage, security review) pass; `audit`/`deny` skipped due to missing binaries in the sandbox only

https://claude.ai/code/session_017vrZ4D421Ty4uKSgmm7Qsb

### DIFF
--- a/examples/lorawan_file_transfer/simulated_radio.rs
+++ b/examples/lorawan_file_transfer/simulated_radio.rs
@@ -1,4 +1,4 @@
-use lora_modulation::{Bandwidth, BaseBandModulationParams, CodingRate, SpreadingFactor};
+use lora_modulation::BaseBandModulationParams;
 use lorawan_device::Timings;
 use lorawan_device::nb_device::radio::{Event, PhyRxTx, Response, RfConfig, RxQuality, TxConfig};
 
@@ -135,6 +135,7 @@ impl Timings for SimulatedRadio {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use lora_modulation::{Bandwidth, CodingRate, SpreadingFactor};
 
     const TEST_SF: u8 = 7;
     const TEST_FREQ: u32 = 868_100_000;
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -321,13 +321,16 @@ impl Channel {
         self.completed
             .iter()
             .filter(|tx| tx.end <= time && !tx.collided)
-            .map(|tx| RxMetadata {
-                payload: tx.payload.clone(),
-                rssi: self.compute_rssi(tx.tx_power_dbm),
-                snr: self.compute_snr(self.compute_rssi(tx.tx_power_dbm)),
-                sf: tx.sf,
-                frequency: tx.frequency,
-                time: tx.end,
+            .map(|tx| {
+                let rssi = self.compute_rssi(tx.tx_power_dbm);
+                RxMetadata {
+                    payload: tx.payload.clone(),
+                    rssi,
+                    snr: self.compute_snr(rssi),
+                    sf: tx.sf,
+                    frequency: tx.frequency,
+                    time: tx.end,
+                }
             })
             .collect()
     }
@@ -736,8 +739,16 @@ mod tests {
         let lora_rx = ch_lora.deliver_to(60_000);
         let strict_rx = ch_strict.deliver_to(60_000);
 
-        assert_eq!(lora_rx.len(), 1, "LoRa threshold=6: strong signal must survive");
-        assert_eq!(strict_rx.len(), 0, "strict threshold=10: both collide at delta=6");
+        assert_eq!(
+            lora_rx.len(),
+            1,
+            "LoRa threshold=6: strong signal must survive"
+        );
+        assert_eq!(
+            strict_rx.len(),
+            0,
+            "strict threshold=10: both collide at delta=6"
+        );
     }
 
     proptest! {
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -224,23 +224,20 @@ impl Scheduler {
                     self.metrics.record_capture();
                 }
                 let mut wakes = Vec::new();
+                let mut tx_node_idxs = Vec::new();
                 for i in 0..self.nodes.len() {
                     if self.nodes[i].node_id() != sender {
+                        let node_id = self.nodes[i].node_id();
                         let next = self.nodes[i].on_receive(frame.clone(), time);
-                        self.metrics.record_rx(self.nodes[i].node_id());
+                        self.metrics.record_rx(node_id);
                         if let Some(t) = next {
-                            wakes.push((self.nodes[i].node_id(), t));
+                            wakes.push((node_id, t));
                         }
+                        tx_node_idxs.push(i);
                     }
                 }
                 for (node_id, t) in wakes {
                     self.schedule(t, EventKind::Wake { node_id });
-                }
-                let mut tx_node_idxs = Vec::new();
-                for i in 0..self.nodes.len() {
-                    if self.nodes[i].node_id() != sender {
-                        tx_node_idxs.push(i);
-                    }
                 }
                 for i in tx_node_idxs {
                     self.handle_poll_transmit(i, time);
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -1,7 +1,7 @@
 use std::cmp::Ordering;
 use std::collections::BinaryHeap;
 
-use crate::channel::{Channel, ChannelConfig, CompletedTx};
+use crate::channel::{Channel, CompletedTx};
 use crate::metrics::MetricsCollector;
 use crate::time::SimTime;
 use crate::traits::InterferenceSource;
--- a/tests/aloha.rs
+++ b/tests/aloha.rs
@@ -329,15 +329,16 @@ fn capture_effect_recorded_in_metrics() {
     // The captured frame is delivered to all non-sender nodes: the weak sender
     // (NodeId(2)) and the receiver (NodeId(99)) — consistent with how the
     // scheduler delivers any successful TX to every non-sender node.
-    assert_eq!(sched.metrics.total_captures, 1, "expected one capture event");
     assert_eq!(
-        sched.metrics.total_collisions,
-        1,
+        sched.metrics.total_captures, 1,
+        "expected one capture event"
+    );
+    assert_eq!(
+        sched.metrics.total_collisions, 1,
         "weak sender should be marked collided"
     );
     assert_eq!(
-        sched.metrics.total_rx,
-        2,
+        sched.metrics.total_rx, 2,
         "captured frame delivered to 2 non-sender nodes (weak sender + receiver)"
     );
 }
--- a/tests/core_coverage.rs
+++ b/tests/core_coverage.rs
@@ -247,8 +247,14 @@ fn channel_late_strong_captures_earlier_weak() {
     ch.resolve_at(80_000);
     let completed = ch.drain_completed();
 
-    let strong_entry = completed.iter().find(|(id, _, _, _)| *id == NodeId(2)).unwrap();
-    let weak_entry = completed.iter().find(|(id, _, _, _)| *id == NodeId(1)).unwrap();
+    let strong_entry = completed
+        .iter()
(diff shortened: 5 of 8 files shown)


DominicBurkart (commented):
Overall the PR does what it says on the tin: seven focused integration tests lock in the invariants described in the PR body, dead code is correctly excised, and the double-RSSI computation is fixed. No blocking correctness issues found. Three things worth addressing before merge, plus a few minor notes.

---

### Issue 1 (medium) — `run_is_bitwise_deterministic_across_runs` is a near-duplicate of `simultaneous_wakes_fire_in_insertion_order`, not an independent determinism check

`run_is_bitwise_deterministic_across_runs` schedules all three nodes at the same single wake time (`Some(100_000)`) and their `update()` returns `None`, so each run consists of exactly one epoch of three simultaneous wakes with no further events. The test passes as long as insertion order is stable between the two `run_once()` calls — which it will always be, because both closures build nodes in the same order. It does not exercise multi-epoch schedules, TX/RX chains, or scenarios where `seq` values span more than one scheduling decision. As a standalone determinism regression-detector it has very low signal density; if someone broke the `seq` tiebreaker for events pushed during a run (e.g. rescheduled wakes from `on_receive`) this test would stay green.

Suggested improvement: have at least one node return `Some(t + delta)` from `update` so the two runs include a second epoch, and/or include a TX/RX pair so events scheduled mid-run are also covered.

---

### Issue 2 (minor) — `wake_past_end_time_is_dropped` asserts `current_time() <= END` but `current_time()` is never advanced

In `Scheduler::run`, `self.current_time = event.time` is only reached *after* the `event.time > self.end_time` guard passes. When the sole event has `time = END + 1`, the guard fires immediately and `current_time` stays at 0. The assertion `sched.current_time() <= END` is therefore vacuously true for any `END >= 0` — it can never fail. A tighter assertion would be `assert_eq!(sched.current_time(), 0)`, which actually encodes the observable postcondition and would catch a future refactor that advances `current_time` before the guard check.

---

### Issue 3 (minor) — `interferer_first_poll_past_end_time_is_dropped` uses `poll_inject` returning `None` but still increments the counter, masking a subtle test gap

`TrackingInterferer::poll_inject` increments `injected` and *then* returns `None`. The assertion `*injected.borrow() == 0` is therefore simultaneously checking two things: (a) the event was never popped from the heap, and (b) the function was never called. These are the same check here because if `poll_inject` is called, `injected` would be 1. However, the intent stated in the PR body is specifically that "a wake at exactly `end_time + 1` is dropped *before dispatch*." The increment-then-return-None pattern means if the scheduler were ever changed to pop-and-filter (advancing `current_time` but not dispatching), the test would still pass while the observable invariant ("the function is never called") would have been violated. Consider splitting the counter increment into a distinct "was called" flag, or keeping the increment but adding an assertion message that clarifies it checks whether `poll_inject` was ever entered, not just whether it injected anything.

---

### Nit — `broadcast_delivers_in_registration_order` silently shares `rx_order` with the sender node

Node id=0 is constructed with `Rc::clone(&rx_order)` as its `pending_rx`. Because the sender never receives its own TX (the `deliver_completed_to_nodes` loop skips `sender`), `rx_order` is never pushed to by node 0 and the assertion is correct. But a reader has to trace through the scheduler source to verify this; a comment noting "sender node shares the Rc but will never push to it because the scheduler skips self-delivery" would make the setup intent clearer.

---

### Nit — `ScheduledEvent::cmp` tiebreaker direction is worth a comment in the source

The ordering is `other.seq.cmp(&self.seq)` — lower seq = higher priority = popped first. This is the correct FIFO direction, but the double-negation (`other` on the left for a max-heap that reverses all comparisons) is easy to misread. The new tests lock in the observable behaviour, but a one-line doc comment on the `cmp` implementation (`// lower seq pops first: FIFO among equal-time events`) would make a future auditor's life easier. Not a test gap, purely a readability note.

---

### Dead-code removal — all correct

- `TwoPhaseProtocol` / `TwoPhaseState` were genuinely unreachable; the replacement rationale comment is accurate and concise.
- `main_binary_is_noop_and_should_be_removed` was an unconditional pass with no assertions; removing it loses no coverage.
- `ChannelConfig` import in `src/scheduler.rs` was used only in the inner `mod tests` which re-imports it locally — clean removal.
- `compute_rssi` double-call in `Channel::deliver_to`: caching the result in `rssi` is correct. The function appears to be pure, so this is a performance fix, not a semantic one — but it's strictly better either way.
- Moving `Bandwidth`/`CodingRate`/`SpreadingFactor` imports into `mod tests` in `simulated_radio.rs` is correct; they were never used outside tests.

---

DominicBurkart:
**CRON status — 2026-04-21** (Claude Code automated run)

Code review posted (3 issues, none blocking — weak determinism probe, vacuously-true assertion in `wake_past_end_time_is_dropped`, conflated "event dropped" vs "function not called" in interferer test). CI is fully green (all checks pass, 100% coverage) but `mergeable_state=dirty` indicating a merge conflict with `main` — this requires manual resolution before the PR can be merged. **Not opened for review** — resolve the merge conflict first, then re-run.

---

DominicBurkart:
**Status (automated orchestration cron, 3-agent fanout)**

Prior DominicBurkart COMMENTED review on head `b3d9498` (at 2026-04-21T11:26:19Z) still stands — near-duplicate determinism test, vacuous `current_time() <= END` assertion, increment-then-return-None masking in interferer test, plus nits on shared `rx_order` and `cmp` tiebreaker direction. No new review was posted (per the since-last-commit guard). Address-comments and CI-fix subagents both hit stream idle timeouts before completing, so nothing new was pushed this cycle. Next cron run should retry both legs.

---

DominicBurkart:
This PR has a `mergeable_state: dirty` conflict in `src/channel.rs` and `src/scheduler.rs` that requires a local `git rebase main` — multiple push attempts via the GitHub API were unable to resolve the underlying git history conflict. Additionally, DominicBurkart's review comments (second-epoch determinism test, interferer counter split, broadcast annotation, cmp doc comment) remain unaddressed. Remaining in draft until both the merge conflict and review comments are resolved. (Session: https://claude.ai/code/session_017vrZ4D421Ty4uKSgmm7Qsb)

---

DominicBurkart (commented):
**Re-review at head `76730b3` (CODE REVIEW subagent, follow-up)**

Only one commit has landed since the prior review at `b3d9498`: a merge-conflict resolution in `src/scheduler.rs` that re-aligns the file with `main`. I re-read the full PR diff and the head `src/scheduler.rs` against `ARCHITECTURE.md`'s determinism promise. All three substantive findings from the 2026-04-21 review still stand unchanged:

1. (medium) `run_is_bitwise_deterministic_across_runs` is a single-epoch, insertion-order-stable scenario — it does not actually probe seq ordering for events scheduled mid-run (e.g. rescheduled wakes from `on_receive`, TX completions). A broken `seq` tiebreaker on dynamically-pushed events would still leave this test green. Extend it to include a second epoch (`update` returning `Some(t + delta)`) and/or a TX/RX exchange.
2. (minor) `wake_past_end_time_is_dropped`: `current_time` is never advanced when the guard fires, so `assert!(current_time() <= END)` is vacuously true. `assert_eq!(current_time(), 0)` encodes the real postcondition (and is in fact what the test already asserts on the second line — the weaker outer assertion is redundant, not harmful).
3. (minor) `interferer_first_poll_past_end_time_is_dropped`: `poll_inject` increments then returns `None`, so the counter conflates "event dispatched" with "function entered." Either split into a `was_called: bool` flag or document that a nonzero counter value catches *any* invocation, injecting or not.

New observations this pass:

- Scheduler refactor in `deliver_completed_to_nodes` (capture `node_id` once, interleave `tx_node_idxs` with the RX walk) is a behaviour-preserving tidy-up: iteration order over `self.nodes` is unchanged, `on_receive` is still called before `handle_poll_transmit` for every non-sender, and wake-scheduling happens between the two phases. The new `scheduler_ordering.rs::broadcast_delivers_in_registration_order` test locks in the observable part of this guarantee.
- No secrets, no oversized additions, dead-code removals remain correct.
- Tests are deterministic (no wall-clock, no thread, no RNG without seed), no flake risk.
- `mergeable_state` is still reported as `dirty` in the PR metadata even after `76730b3`; worth a `git fetch && git merge-base --is-ancestor origin/main HEAD` check locally before requesting review again.

Recommendation: address the three outstanding items (especially #1 — it's the only one with real regression-detection impact), rebase cleanly onto `main`, then move out of draft. No blocking correctness concerns.

https://claude.ai/code/session_01PATZg9FPFycvHzZoed9oiv

---

PR #63, opened 2026-04-24
test(scheduler): contract tests for InterferenceSource::observe dispatch

DominicBurkart:
## Gap

`InterferenceSource::observe` is the sole mechanism by which interferers learn about activity on the shared radio medium (any `ChannelEvent` — `TransmissionStarted` / `TransmissionCompleted`). The scheduler dispatches these to every registered interferer from three call-sites in `src/scheduler.rs`:

1. `handle_poll_transmit` (node TX start) — `interferer.observe(&ch_event, time)`
2. `TxComplete` resolution (TX completion) — `interferer.observe(ch_event, event.time)`
3. `InterferencePoll` (interferer-originated TX start) — `self.interferers[i].observe(&ch_event, time)`

## Prior disposition

Prior to this PR, this dispatch path had **zero** assertions. A search across `tests/` and `src/` shows every existing interferer implementation simply `_`-ignores the `event` argument to `observe`:

```rust
fn observe(&mut self, _event: &ChannelEvent, _time: SimTime) {}
```

A silent regression (skipped dispatch, dropped `TransmissionCompleted` for interferer-originated TXs, notifying only the injecting interferer, etc.) would pass every existing test while silently breaking every real interferer-dependent protocol (adaptive jammers, carrier-sense models, passive traffic monitors). Only the side-effect counters `total_airtime_us` / `total_collisions` indirectly observed interferer TXs — nothing verified the event stream itself.

## Invariants

`tests/interferer_observe_contract.rs` (8 cases) locks down the following contract:

- **Delivery**: every node TX produces one `TransmissionStarted` and one `TransmissionCompleted` at every registered interferer.
- **Timing fidelity**: the `time` field on every observed `ChannelEvent` equals the `time` argument the scheduler passed to `observe` (no drift between the scheduler clock and event payload).
- **Started/Completed balance**: per-sender started-count equals completed-count across a run that completes before end-time.
- **Broadcast**: all registered interferers — not just one — receive the full event sequence.
- **Self-observation**: an injecting interferer observes its own TX's Started and Completed events.
- **Peer observation**: sibling interferers observe a peer's injection's Started and Completed events.
- **Ordering**: `Started` strictly precedes `Completed` for the same sender, with a gap equal to the TX duration.
- **Collision preservation**: collided TXs still emit both events, with `collided: true` on the `Completed` payload.

## Verification

- `cargo test --test interferer_observe_contract` → 8 passed, 0 failed.
- `cargo clippy --all-features --all-targets -- -D warnings` → clean.
- `cargo fmt --check tests/interferer_observe_contract.rs` → clean.

https://claude.ai/code/session_01LEodkCtiog4uAeXmJwHnWB

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -51,7 +51,7 @@ struct Transmission {
 }
 ```
 
-`init`, `on_receive`, and `update` each return `Option<SimTime>` — the next simulation time at which the scheduler must call `update` on this node. Returning `None` means the node has no pending timer. This allows the scheduler to use event-driven dispatch (a priority queue keyed on SimTime) rather than polling every node on every tick.
+`init`, `on_receive`, and `update` each return `Option<SimTime>` — the next simulation time at which the scheduler must call `update` on this node, or `None` if the node has no pending timer. This lets the scheduler use event-driven dispatch (a priority queue keyed on SimTime) instead of polling every node on every tick.
 
 `update` drives timer-based state transitions (e.g. RX1/RX2 window opening in LoRaWAN Class A) without requiring an incoming frame.
 
@@ -105,7 +105,7 @@ The scheduler calls `poll_transmit` on the protocol; the protocol (or its adapte
 
 ### Validation Target: LoRaWAN via lora-rs
 
-LoRaWAN via lora-rs is the first real-world protocol used to prove the simulation engine works with a real stack. The validation example is external to theatron core and comprises three components:
+LoRaWAN via lora-rs is the first real-world protocol used to validate the simulation engine against a production stack. The validation example is external to theatron core and comprises three components:
 
 - **LoRaWAN device adapter**: wraps `lorawan-device::nb_device` to implement the `Protocol` trait
 - **Simulated network server**: responds to joins and schedules downlinks (see below)
--- a/examples/aloha/aloha_node.rs
+++ b/examples/aloha/aloha_node.rs
@@ -3,16 +3,16 @@ use theatron::time::SimTime;
 use theatron::traits::TrafficModel;
 use theatron::types::{NodeId, RxMetadata, Transmission};
 
-// Default LoRa EU868 radio parameters. These are standard values used across
-// all AlohaNode instances. `sf` and `frequency` remain caller-supplied because
-// they are the primary parameters for channel / orthogonality experiments.
-// TODO: consider accepting all RF params via an AlohaConfig struct and
-// implementing the `Protocol` trait (per ARCHITECTURE.md) so that this example
-// demonstrates the idiomatic theatron integration pattern rather than wiring
-// directly to `NodeHandle`.
+// Default LoRa EU868 radio parameters used across all AlohaNode instances.
+// `sf` and `frequency` are caller-supplied because they are the primary
+// parameters for channel and orthogonality experiments.
+//
+// TODO: accept all RF params via an `AlohaConfig` and implement the `Protocol`
+// trait (see ARCHITECTURE.md) so this example demonstrates the idiomatic
+// integration pattern rather than wiring directly to `NodeHandle`.
 const DEFAULT_BANDWIDTH_HZ: u32 = 125_000; // EU868 standard bandwidth
 const DEFAULT_CODING_RATE: u8 = 5; // 4/5 coding rate
-const DEFAULT_TX_POWER_DBM: i8 = 14; // 14 dBm, legal limit for EU868
+const DEFAULT_TX_POWER_DBM: i8 = 14; // EU868 legal limit
 
 /// Pure ALOHA node: transmits immediately when a payload is available.
 ///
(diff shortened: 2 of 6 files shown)


DominicBurkart (commented):
## Review: contract tests for `InterferenceSource::observe` dispatch

Overall this is a solid, well-structured test file that fills a genuine gap in coverage. The module-level doc comment clearly explains the contract, the helpers are clean, and all 8 tests are deterministic (no threads, no OS timing) so there is no flakiness risk. The -62 deletions are purely comment cleanup — no logic was touched.

A few issues worth addressing before merging, ranging from a hard-coded internal assumption to missing coverage.

---

### Hard-coded synthetic `NodeId` is a brittle coupling

`interferer_observes_own_injected_tx_start_and_complete` and `sibling_interferer_observes_peer_injected_tx` both rely on `NodeId(u32::MAX)` being the scheduler-assigned ID for the first registered interferer:

```rust
// The interferer is registered as synthetic NodeId(u32::MAX).
let synthetic = NodeId(u32::MAX);
```

If the scheduler changes its internal ID-assignment scheme (e.g., to allocate from a counter or to start from a different sentinel), these tests will immediately fail with a confusing `0 != 1` message rather than a clear "wrong ID" message. More importantly, the assumption is invisible at the call site — a reader of the scheduler source has to know to look here.

**Suggestion:** expose the synthetic ID assignment through the `add_interferer` return value (returning the assigned `NodeId`), or expose a `const INTERFERER_BASE_ID: NodeId` sentinel. If neither is acceptable right now, at minimum assert the *total* event count in addition to the per-sender count so the test is falsifiable even when the sender ID changes.

---

### `sibling_interferer_observes_peer_injected_tx` — unused comment, incomplete verification

The comment "Second interferer is synthetic `NodeId(u32::MAX - 1)`" names a value that is never used in the test body. The peer log is never checked for *absence* of its own (non-existent) TX events either. The test correctly checks that the peer sees 1 started + 1 completed from the injector, but it would be stronger to also assert that `peer_log.len() == 2` (exactly those two events, nothing extra).

---

### `all_registered_interferers_receive_same_events` — timing match is necessary but not sufficient

The test checks that all three logs have the same per-index `.time`, which is good. But it does not check that the *event variants and payloads* are identical across logs. A regression that delivers a `TransmissionStarted` to log_a but a `TransmissionCompleted` to log_b at the same timestamp would pass the current assertions.

**Suggestion:** add a structural equality check, e.g.:

```rust
for i in 0..2 {
    assert_eq!(a[i].time, b[i].time);
    assert_eq!(b[i].time, c[i].time);
    // also verify same event kind:
    assert!(matches!((&a[i].event, &b[i].event),
        (ChannelEvent::TransmissionStarted { .. }, ChannelEvent::TransmissionStarted { .. }) |
        (ChannelEvent::TransmissionCompleted { .. }, ChannelEvent::TransmissionCompleted { .. })
    ));
}
```

Or, if `ChannelEvent` derives `PartialEq`, a direct `assert_eq!(a[i].event, b[i].event)` is simpler.

---

### Missing scenario: node TX + interferer TX coexist

There is no test for the case where both a node and an interferer inject transmissions in the same simulation run and both registrations must observe *each other's* events. This is the most realistic scenario (an adaptive jammer reacting to node traffic) and the one most likely to expose a bug where the scheduler only fans out events to interferers that are not the source of a given TX. The existing tests cover node-only and interferer-only scenarios but not the mixed case.

---

### Minor: `LoggingActiveInterferer::next_poll_time` returns a future time after exhaustion in the same tick

```rust
fn next_poll_time(&self, current_time: SimTime) -> Option<SimTime> {
    if self.remaining > 0 {
        Some(current_time + self.interval)
    } else {
        None
    }
}
```

Because `poll_inject` decrements `remaining` before the scheduler calls `next_poll_time` again, this is safe in practice. But if the scheduler's event loop ever calls `next_poll_time` *before* `poll_inject` within the same step, it would schedule a spurious second poll that would harmlessly return `None`. No action needed unless you want the helper to be reusable outside this file.

---

### Non-issues confirmed

- **Deletions (-62):** All comment cleanup in `protocol_timer_contract.rs`, `aloha.rs`, `aloha_node.rs`, and `core_coverage.rs`. No logic changed.
- **Flakiness:** Zero — pure deterministic event-driven simulation, no system clock, no concurrency.
- **Test isolation:** Each test constructs its own `Scheduler` and `Rc<RefCell<_>>` log; no shared mutable state across tests.
- **`observe_event_time_matches_dispatch_time`:** Correctly uses a non-zero start offset (`Some(10_000)`) which would catch an off-by-one if the scheduler passed `t=0` to `observe` regardless of actual TX time.
- **`collided_tx_still_emits_started_and_completed`:** Collision setup is sound — SF=7, same frequency, node 2 starts at t=10_000 while node 1's 50_000 µs TX is still in progress, so overlap is guaranteed.

The main asks before merge are: (1) address the `NodeId(u32::MAX)` coupling, (2) add the mixed node+interferer scenario, and (3) strengthen the broadcast test to compare event payloads not just timestamps.

---

PR #83, opened 2026-04-29
refactor(lorawan_adapter): remove dead `joined` field

DominicBurkart:
## Summary

`LoRaWanAdapter::joined: bool` is initialised to `true` in `new()` and never mutated again. The only read site is the early-exit guard

```rust
if !self.joined || !self.device.ready_to_send_data() {
    return self.fragmenter.next_available_time(time);
}
```

so `!self.joined` is a permanently-false short-circuit. The field, the initialiser entry, and the dead `||` branch are removed; the rest of `try_send_fragment` is unchanged.

## Why

ARCHITECTURE.md's "Adapter state ownership" section says `Protocol::State` should only carry "bookkeeping theatron needs" — `joined` carries no information at runtime. The change also fits the broader `arch/reduce-redundancy` work stream that targets dead state in the codebase.

## Scope

One file (`examples/lorawan_file_transfer/lorawan_adapter.rs`), one logical change. No behaviour change: `device.ready_to_send_data()` already returns `true` only after the ABP join completes, which is exactly what the dead bool was tracking.

## Test plan
- [x] `cargo check --all-features --all-targets`
- [x] `cargo clippy --all-features --all-targets -- -D warnings`
- [x] `cargo test --test lorawan_file_transfer` (6/6 passed)
- [x] `cargo test --example lorawan_file_transfer` (18/18 passed)

https://claude.ai/code/session_01C5q6TXtM8gokDqpRTMWRe5

---
_Generated by [Claude Code](https://claude.ai/code/session_01C5q6TXtM8gokDqpRTMWRe5)_

### DIFF
--- a/examples/lorawan_file_transfer/lorawan_adapter.rs
+++ b/examples/lorawan_file_transfer/lorawan_adapter.rs
@@ -19,7 +19,6 @@ pub struct LoRaWanAdapter {
     fragmenter: FileFragmenter,
     pending_timeout_ms: Option<u32>,
     tx_start_time: SimTime,
-    joined: bool,
     /// Transmission staged during `on_receive` / `update` so that
     /// `poll_transmit` never needs to reach into the radio directly.
     pending_tx: Option<Transmission>,
@@ -46,7 +45,6 @@ impl LoRaWanAdapter {
             fragmenter,
             pending_timeout_ms: None,
             tx_start_time: 0,
-            joined: true,
             pending_tx: None,
         }
     }
@@ -65,7 +63,7 @@ impl LoRaWanAdapter {
     }
 
     fn try_send_fragment(&mut self, time: SimTime) -> Option<SimTime> {
-        if !self.joined || !self.device.ready_to_send_data() {
+        if !self.device.ready_to_send_data() {
             return self.fragmenter.next_available_time(time);
         }
         if let Some(payload) = self.fragmenter.next_payload(time) {


DominicBurkart (commented):
**Automated code review — SHA 780d99ee**

The change is correct. All three claims in the PR summary check out.

**`joined` was always `true`**
The field was set to `true` in `new()` and never mutated anywhere in the file. The `!self.joined` short-circuit in `try_send_fragment` was therefore always `false` and could never fire — pure dead code.

**`device.ready_to_send_data()` fully covers the join semantics here**
The adapter uses `JoinMode::ABP` exclusively. ABP provisioning is synchronous and infallible (`device.join(credentials).expect("ABP join must succeed")`), so the device is joined before the struct is returned from `new()`. For ABP, `ready_to_send_data()` is strictly more informative than the static `joined` flag: it reflects the live MAC state (not transmitting, not in RX windows, etc.) rather than a one-time boolean that could never become `false`. The two checks are not equivalent in general, but for this ABP-only constructor they are, and `ready_to_send_data()` is the correct long-term gate.

**No residual `joined` references**
The diff removes all three occurrences: the struct field declaration, the field initialization, and the guard in `try_send_fragment`. Nothing left to clean up.

**ARCHITECTURE.md consistency**
The "Adapter state ownership" section confirms `nb_device::Device` encapsulates full MAC state, and the Class A state flow diagram shows idle-state gating belongs to the device's own state machine — consistent with delegating to `ready_to_send_data()`.

**One forward-looking note**: if this example is ever extended to support OTAA in addition to ABP, `ready_to_send_data()` still guards correctly (it returns `false` until the join-accept exchange completes), so no structural change would be needed even then.

**Verdict: LGTM, safe to merge.**

---

DominicBurkart:
Review confirmed `joined` was always true and `device.ready_to_send_data()` fully covers the ABP join semantics; ARCHITECTURE.md's "Adapter state ownership" section is consistent with the removal. No review comments to address. All 6 CI check runs pass with 100% coverage and mergeable state is clean — opening for human review now. [Session](https://claude.ai/code/session_01C26t1HxUSdwQMCqmijYXeq)

---

PR #84, opened 2026-04-29
refactor: drop redundant `ANTENNA_GAIN = 0` from PhyRxTx impls

DominicBurkart:
## Summary

`lorawan_device::nb_device::radio::PhyRxTx` declares `ANTENNA_GAIN` with a default value of `0` (see `lorawan-device/src/nb_device/radio.rs:37` in the pinned upstream rev). ARCHITECTURE.md's "PhyRxTx — the actual interface" block also documents the default as `0`. Both of our impls

- `examples/lorawan_file_transfer/simulated_radio.rs` (`SimulatedRadio`)
- `tests/lorawan_frame_correctness.rs` (`MinimalRadio`)

re-state `const ANTENNA_GAIN: i8 = 0;`, which is identical to the trait default and therefore pure boilerplate. The two lines are removed.

## Why

Aligns with the project's existing `arch/reduce-redundancy` work stream by deleting state that carries no information beyond the upstream trait contract. Behaviour is identical: the impls fall through to the same default value.

## Scope

Two single-line deletions. No public API change, no behaviour change. `MAX_RADIO_POWER` is kept because it has no trait default — each impl must specify it.

## Test plan
- [x] `cargo check --all-features --all-targets`
- [x] `cargo clippy --all-features --all-targets -- -D warnings`
- [x] `cargo test --test lorawan_frame_correctness` (3/3 passed)
- [x] `cargo test --example lorawan_file_transfer` (18/18 passed)

https://claude.ai/code/session_01C5q6TXtM8gokDqpRTMWRe5

---
_Generated by [Claude Code](https://claude.ai/code/session_01C5q6TXtM8gokDqpRTMWRe5)_

PR #88, opened 2026-04-30
arch: replace pending_downlink Vec with bool flag in SimulatedRadio

DominicBurkart:
## Summary

`SimulatedRadio` carried `pending_downlink: Option<Vec<u8>>`. The Vec was set in `inject_downlink` to `Some(data)` and consumed only via `pending_downlink.take().is_some()` inside `Event::Phy`. The bytes themselves were never read — payload bytes already live in `rx_buf`, which is what `get_received_packet` returns to the upstream `lorawan-device` state machine. The `Option<Vec<u8>>` was therefore a heap-allocated boolean.

## Changes

- Replace `pending_downlink: Option<Vec<u8>>` with `downlink_ready: bool`.
- `Event::Phy` consumes the flag via `std::mem::take`.
- Drop the dead-coded `has_pending_downlink` helper (annotated `#[allow(dead_code)]`).
- Replace the `new_radio_no_pending_downlink` test (which only exercised the dropped helper) with a test that asserts the flag is consumed one-shot.

## Why

ARCHITECTURE.md > Validation Target > SimulatedRadio sketch describes the radio as maintaining "an internal receive buffer and an RX-mode flag" — the production code carried both that buffer and a redundant Vec carrying the same bytes. This PR brings the implementation back in line with the sketch (a flag plus the buffer) and removes one heap allocation per simulated downlink.

## Precept tied

ARCHITECTURE.md > SimulatedRadio sketch: "an internal receive buffer and an RX-mode flag." The dead Vec contradicted the documented design.

Reduction-of-redundancy: the value carried by `Option<Vec<u8>>` was indistinguishable from `bool`.

## Test plan

- [ ] `cargo test --example lorawan_file_transfer`
- [ ] `cargo test --test lorawan_file_transfer`
- [ ] `cargo clippy --all-targets -- -D warnings`

https://claude.ai/code/session_ecstatic-planck-Ed9Km

---
_Generated by [Claude Code](https://claude.ai/code/session_01P4jqYEaszguJ91VUxtR3XV)_

### DIFF
--- a/examples/lorawan_file_transfer/simulated_radio.rs
+++ b/examples/lorawan_file_transfer/simulated_radio.rs
@@ -12,7 +12,7 @@ pub struct SimulatedRadio {
     rx_buf: [u8; 256],
     rx_len: usize,
     pending_tx: Option<Transmission>,
-    pending_downlink: Option<Vec<u8>>,
+    downlink_ready: bool,
     current_rx_config: Option<RfConfig>,
 }
 
@@ -22,7 +22,7 @@ impl SimulatedRadio {
             rx_buf: [0u8; 256],
             rx_len: 0,
             pending_tx: None,
-            pending_downlink: None,
+            downlink_ready: false,
             current_rx_config: None,
         }
     }
@@ -44,13 +44,8 @@ impl SimulatedRadio {
         let len = data.len().min(256);
         self.rx_buf[..len].copy_from_slice(&data[..len]);
         self.rx_len = len;
-        self.pending_downlink = Some(data);
+        self.downlink_ready = true;
         true
-    }
-
-    #[allow(dead_code)]
-    pub fn has_pending_downlink(&self) -> bool {
-        self.pending_downlink.is_some()
     }
 }
 
@@ -112,7 +107,7 @@ impl PhyRxTx for SimulatedRadio {
                 Ok(Response::Idle)
             }
             Event::Phy(()) => {
-                if self.pending_downlink.take().is_some() {
+                if std::mem::take(&mut self.downlink_ready) {
                     Ok(Response::RxDone(RxQuality::new(-80, 10)))
                 } else {
                     Ok(Response::Idle)
@@ -159,9 +154,9 @@ mod tests {
     }
 
     #[test]
-    fn new_radio_no_pending_downlink() {
+    fn new_radio_no_downlink_ready() {
         let radio = SimulatedRadio::new();
-        assert!(!radio.has_pending_downlink());
+        assert!(!radio.downlink_ready);
     }
 
     #[test]
@@ -218,6 +213,17 @@ mod tests {
     #[test]
     fn phy_without_downlink_returns_idle() {
         let mut radio = SimulatedRadio::new();
+        let result = radio.handle_event(Event::Phy(()));
+        assert!(matches!(result, Ok(Response::Idle)));
+    }
+
+    #[test]
+    fn phy_consumes_downlink_flag() {
+        let mut radio = SimulatedRadio::new();
+        let _ = radio.handle_event(Event::RxRequest(make_rf()));
+        assert!(radio.inject_downlink(vec![0xAB], TEST_SF, TEST_FREQ));
+        let _ = radio.handle_event(Event::Phy(()));
+        // Second Phy event must report Idle: the flag is one-shot.
         let result = radio.handle_event(Event::Phy(()));
         assert!(matches!(result, Ok(Response::Idle)));
     }
--- a/tests/lorawan_frame_correctness.rs
+++ b/tests/lorawan_frame_correctness.rs
@@ -12,7 +12,7 @@ struct MinimalRadio {
     rx_buf: [u8; 256],
     rx_len: usize,
     pending_tx: Option<Transmission>,
-    pending_downlink: Option<Vec<u8>>,
+    downlink_ready: bool,
 }
 
 impl MinimalRadio {
@@ -21,7 +21,7 @@ impl MinimalRadio {
             rx_buf: [0u8; 256],
             rx_len: 0,
             pending_tx: None,
-            pending_downlink: None,
+            downlink_ready: false,
         }
     }
 
@@ -72,7 +72,7 @@ impl PhyRxTx for MinimalRadio {
             Event::RxRequest(_) => Ok(Response::Rxing),
             Event::CancelRx => Ok(Response::Idle),
             Event::Phy(()) => {
-                if self.pending_downlink.take().is_some() {
+                if std::mem::take(&mut self.downlink_ready) {
                     Ok(Response::RxDone(RxQuality::new(-80, 10)))
                 } else {
                     Ok(Response::Idle)


DominicBurkart (commented):
The core refactor is correct and the reasoning holds up under inspection. Detailed findings below.

---

**What checks out**

- **Vec bytes were never read — confirmed.** In the pre-PR code, `inject_downlink` immediately copies into `rx_buf`/`rx_len` and then stored the same `data` as `pending_downlink = Some(data)`. The only downstream read of that field was `pending_downlink.take().is_some()`. The `Vec` content was never accessed; `rx_buf` is the authoritative payload buffer. The refactor is semantically exact.

- **`std::mem::take` on `bool` is correct.** `bool`'s `Default` is `false`, so `std::mem::take(&mut self.downlink_ready)` returns the current value and resets the field to `false` in one step — exactly the one-shot consume pattern intended.

- **`phy_consumes_downlink_flag` tests the right thing.** It injects a downlink, fires `Event::Phy(())` once (consuming the flag), then fires a second `Event::Phy(())` and asserts `Response::Idle`. This directly validates the one-shot guarantee the PR description promises.

- **All call-sites inside `simulated_radio.rs` are updated.** No references to `pending_downlink` or `has_pending_downlink` remain in that file.

- **Codecov reports 100% coverage maintained.**

---

**Issue that should be fixed before merge**

`tests/lorawan_frame_correctness.rs` contains a separate `MinimalRadio` struct that was not touched by this PR and still carries the identical dead-Vec pattern:

```rust
// tests/lorawan_frame_correctness.rs  (unchanged on this branch)
struct MinimalRadio {
    // ...
    pending_downlink: Option<Vec<u8>>,
}

// in handle_event:
Event::Phy(()) => {
    if self.pending_downlink.take().is_some() {
        Ok(Response::RxDone(RxQuality::new(-80, 10)))
    } else {
        Ok(Response::Idle)
    }
}
```

`MinimalRadio` is a different type from `SimulatedRadio`, so this file still compiles cleanly and no existing test breaks. But the PR's stated rationale — "a heap-allocated boolean" / "remove one heap allocation per simulated downlink" / "brings the implementation back in line with the ARCHITECTURE.md sketch" — applies to `MinimalRadio` equally. The PR claims to sweep the codebase clean of this pattern; `MinimalRadio` is left behind.

This should be updated to `downlink_ready: bool` / `std::mem::take` (mirroring exactly what was done for `SimulatedRadio`) before merge, so the codebase is internally consistent and the PR's stated goal is fully achieved.

---

**Minor observation (no change required)**

`new_radio_no_downlink_ready` reads the private field directly (`assert!(!radio.downlink_ready)`). This works because the test lives in the same file's inline `mod tests` block and is idiomatic Rust. A behavioral assertion via `handle_event(Event::Phy(()))` returning `Idle` would be equally valid and slightly more black-box, but this is a style preference, not a defect.

---

DominicBurkart:
**CRON review status:** The `Option<Vec<u8>>` → `bool` refactor is correct — the Vec bytes were verified as never read (only `.is_some()` was checked), `std::mem::take` on `bool` is sound (Default is `false`), and the one-shot consumption test correctly covers the invariant. A fix was pushed to also update `MinimalRadio` in `tests/lorawan_frame_correctness.rs`, which still carried the old pattern and contradicted the PR's stated goal. This PR is now open for human review; CI was green before the fix and the test-file-only update should continue to pass.

https://claude.ai/code/session_017X42UqJxgKM5zvE2VjgN84

---

DominicBurkart (commented):
Re-review on head `ed8d24e` (audience: coding agents). The prior review's blocker is resolved and the refactor is internally consistent.

---

**Vec to bool replacement: correctness verified**

- **No surviving readers of the Vec contents.** `search_code` for `pending_downlink` across the repo returns zero hits. The only pre-PR consumer was `pending_downlink.take().is_some()` in the `Event::Phy` arm; the bytes were never read because `rx_buf` (a fixed `[u8; 256]`) is the authoritative payload buffer that backs `get_received_packet`. Replacing `Option<Vec<u8>>` with `bool` is information-preserving.

- **No `.len()` / iteration / queue-depth use.** A `bool` cannot represent a count, but no code ever asked for one. `inject_downlink` already overwrites `rx_buf` and `rx_len` unconditionally, so the prior data structure was never a queue — at most one downlink could be represented at any time, regardless of `Option<Vec<u8>>` vs `bool`. The new type matches the actual semantics ("one pending downlink at a time").

- **`std::mem::take(&mut self.downlink_ready)` is equivalent to the prior `.take().is_some()`.** `bool::default()` is `false`, so the expression returns the current value and resets the field in a single atomic-from-the-caller's-view step. The `Event::Phy` branch still emits `RxDone` exactly when a downlink was injected since the last consume, and `Idle` otherwise. No ordering, drop-if-pending, or queue-depth behavior change.

- **`MinimalRadio` in `tests/lorawan_frame_correctness.rs` is migrated.** This addresses the only blocker from the prior review on `18aa931`. The codebase no longer carries the dead-Vec pattern in any form.

---

**Test coverage of both transitions**

- `new_radio_no_downlink_ready`: initial-clear assertion (white-box, but acceptable for a same-file `mod tests`).
- `phy_without_downlink_returns_idle`: clear → `Idle`.
- `phy_with_downlink_returns_rx_done`: set → consume → `RxDone`.
- `phy_consumes_downlink_flag` (new): set → consume → second consume returns `Idle`. This is the load-bearing one-shot guarantee and is now explicit. Good.
- `inject_downlink_populates_rx_buf` continues to assert the buffer is the source of truth, which is what justifies dropping the Vec.

The integration tests in `tests/lorawan_frame_correctness.rs` (`uplink_frame_parses_as_valid_lorawan`, `uplink_mic_validates`, `frame_counter_increments`) exercise the TX path through the migrated `MinimalRadio` and would fail to compile if the rename were inconsistent.

---

**No issues found that block merge.**

Minor nit (no change required, same as prior review): `new_radio_no_downlink_ready` reads the private field directly. A behavioral check via `Event::Phy(()) -> Idle` would be slightly more black-box, but the in-file `mod tests` access is idiomatic Rust and not worth churning.

Recommendation: **ready_for_human**.

---
_Generated by Claude Code review subagent_

---

DominicBurkart:
## CRON status

Posted review confirming the Vec→bool replacement is information-preserving (no `.len()`/iteration/queue-depth readers existed; `rx_buf` was always the authoritative payload), `std::mem::take` correctly preserves one-shot semantics, and `MinimalRadio` is migrated alongside `SimulatedRadio` (resolving the prior review's blocker); no actionable comments. All 6 GitHub Actions jobs passed; the only CI red is `codecov/project` flagging −0.48% (sibling refactor PRs #84/#86/#87 hit the same threshold artifact from deletion of covered lines), which is a repo policy decision rather than a code defect. Branch was updated against latest base via `update_pull_request_branch`; staying draft pending the fresh CI run + owner decision on the codecov threshold. — [session](https://claude.ai/code/session_98067ac6-32fa-42aa-af0b-967b32eaf819)

---

PR #89, opened 2026-04-30
test: cover InterferenceSource::observe scheduler contract

DominicBurkart:
## Component

`InterferenceSource` integration with `Scheduler` (`src/traits.rs`, `src/scheduler.rs`).

## Current disposition

`InterferenceSource` is the extension point for adaptive jammers, energy-detection / CSMA-style interferers, and any traffic source that wants to *react* to channel activity. The scheduler is the only path through which an `InterferenceSource` learns about channel events: it calls `observe()` from three places (node TX start in `handle_poll_transmit`, interferer TX start in `EventKind::InterferencePoll`, and on every completion in `EventKind::TxComplete`), and it polls injection via `poll_inject()` / `next_poll_time()`.

Existing tests cover *that* an interferer can be registered and that an active injector adds airtime / collides with a node, but they never inspect the events delivered to `observe()`. A regression in the notification fan-out (e.g. dropping a self-observation, missing a completion event, or losing the `collided` flag) would silently break every adaptive interference model without any existing test catching it.

## Invariants tested

1. **Bidirectional observation** — every node TX produces both a `TransmissionStarted` (at `start_time`) and a `TransmissionCompleted` (at `end_time`) event delivered to every registered interferer, with sender / sf / frequency / time matching the originating `Transmission`.
2. **Self-observation** — an interferer that injects a TX via `poll_inject` also receives the resulting `TransmissionStarted` (under the synthetic `NodeId(u32::MAX - idx)`) and the matching `TransmissionCompleted`.
3. **Fan-out across interferers** — when interferer A injects a TX, interferer B observes it; verifies the loop in the scheduler is over *all* interferers, not just the originator.
4. **Faithful `collided` propagation** — overlapping same-SF/freq node TXs are reported to interferers with `collided=true` on both completion events, so adaptive sources can distinguish successful TXs from collisions.
5. **Polling-stop contract** — once `next_poll_time` returns `None`, the scheduler must never call `poll_inject` again (no spurious re-scheduling, no re-injection from a one-shot source).
6. **Temporal monotonicity of observations** — across a run, the times passed to `observe` are non-decreasing, the basic invariant time-windowed adaptive logic depends on.

## Strategy

**Integration tests** end-to-end through `Scheduler::run`. Doctests / unit tests on the trait alone could not exercise the scheduler's notification fan-out (the actual locus of the contract), and proptests would buy little here because each invariant is a discrete event-routing assertion rather than a numeric property. The fixtures use `Rc<RefCell<...>>` to capture observations from inside `&mut self` callbacks without taking ownership back from the scheduler — no sleeps, no timing dependence, fully deterministic.

## Files

- New: `tests/interferer_observe_contract.rs`

## Test plan

- [ ] CI green on this branch
- [ ] No flakiness / no sleeps / no mock-only assertions

https://claude.ai/code/session_01P4jqYEaszguJ91VUxtR3XV

---
_Generated by [Claude Code](https://claude.ai/code/session_01P4jqYEaszguJ91VUxtR3XV)_


DominicBurkart (commented):
Reviewed `tests/interferer_observe_contract.rs` against `src/scheduler.rs` and `src/traits.rs` at head commit `db3bd43`.

**Overall:** The tests are well-structured, fully deterministic, and the invariants map cleanly onto the scheduler's actual code paths. No borrow-panic risk. CI is green at 100% coverage. A few correctness notes and one genuine gap worth addressing before marking ready.

---

## Correctness

### `next_poll_time_none_permanently_stops_polling` — assertion proves less than claimed

The test asserts `*poll_count == 3`. That passes because the implementation happens to stop, but it doesn't prove the *permanent-stop* invariant directly. If the scheduler had a latent bug where it re-polled after a long quiet period (e.g. after a late-arriving `TxComplete` event re-queued the interferer), the count would exceed 3 and the assertion would still catch it — so in practice this is fine. However, the assertion message says "must be called exactly max_polls times once `next_poll_time` returns `None`", which is accurate only because `max_polls=3` and no other mechanism can re-schedule. A brief comment clarifying *why* equality is the right check (vs. `<=`) would help future readers.

### `interferer_observes_its_own_injected_tx` — `TransmissionCompleted` sender unchecked

The test verifies `TransmissionStarted.sender == NodeId(u32::MAX)` but never checks the corresponding `TransmissionCompleted` sender. Completing that check would close the "self-observation also covers completion" half of invariant 2.

```rust
if let ChannelEvent::TransmissionCompleted { sender, .. } = &completes[0].0 {
    assert_eq!(*sender, NodeId(u32::MAX));
}
```

### `interferer_observes_node_tx_start_and_completion` — completion `time` field vs. second argument to `observe`

The test destructs `ChannelEvent::TransmissionCompleted { time, .. }` and asserts `*time == 50_000`. This tests the value inside the event struct, not the `SimTime` argument passed to `observe`. The scheduler passes `event.time` as the second argument, which equals 50_000 here. Both values *should* agree (and Codecov shows they do), but if `Channel::resolve_at` ever sets the event's `time` field to a different value (e.g. 0 or the start time), this test would not catch that divergence. If the `ChannelEvent::TransmissionCompleted.time` field is meant to be the completion time, the assertion is testing the right thing; it is just worth a comment that both paths are being exercised.

---

## Borrow-safety

All `Rc<RefCell<...>>` usages are safe:

- `observe` (`borrow_mut`) and `next_poll_time` (`borrow`) are never called concurrently — the scheduler calls `poll_inject` to completion, then `next_poll_time`, then the `observe` loop for a separate event.
- The test body only borrows the `Rc<RefCell<...>>` after `sched.run()` returns, so there is no overlap with any in-flight `borrow_mut` inside the scheduler.
- `InjectingObserver.inject` is a plain `RefCell` (not `Rc`); `borrow_mut().take()` inside `poll_inject` is called once and finishes before any `observe` call.

No runtime borrow panics are possible in the current test structure.

---

## Edge-case gaps

### `each_interferer_observes_other_interferers_tx` — count-1 assertion is ambiguous

A observes its own TX (self-observation) and B observes A's TX (fan-out). Both produce 1 start + 1 complete, so the assertion passes for both. But A's count of 1 is consistent with *either* self-observation working correctly *or* the fan-out working correctly — you cannot tell from the count alone whether A received its own event or B's.

Since there's only one TX in this test (A injects, B is silent), the counts are unambiguous in practice. But if you want this test to specifically target fan-out rather than self-observation, it would be cleaner to use a node TX (not an interferer injection) as the event source, so A's observation is clearly "receiving a foreign TX" rather than self-observation. Alternatively, a comment noting the distinction would suffice.

### Collision test uses `Some(10_000)` for node 2's wake — overlap is assumed, not guaranteed

`interferer_observes_collided_flag_on_completion` wakes node 2 at t=10_000, while node 1 TX runs from t=0 to t=50_000. Node 2's TX also runs for 50_000 µs, so it overlaps [10_000, 60_000] ∩ [0, 50_000] = [10_000, 50_000]. The collision is real and deterministic here. But the scheduler `end_time` is set to 500_000 in that test — the collision completes at t=60_000 (node 2 ends), well within bounds. This is fine; just worth noting that changing `duration_us` or wake times could silently break the test's collision assumption without any compile-time protection.

---

## Documentation mismatch

The PR description lists **6 invariants**; the file-level doc comment lists only **5** (the fan-out invariant — "interferer A injects, interferer B observes it" — is elided from the module doc). Minor, but mismatches like this can confuse future readers about what is covered.

---

## Summary

| Finding | Severity |
|---|---|
| `TransmissionCompleted` sender unchecked in self-observation test | Low — missing assertion |
| Count-based assertion in polling-stop test is correct but underexplained | Cosmetic |
| Fan-out test count-1 assertion is ambiguous between self-observation and fan-out | Low — functionally correct |
| Module doc lists 5 invariants, PR description says 6 | Cosmetic |
| Borrow safety | No issues |
| Compile correctness (traits, types, method signatures) | No issues |

The most actionable fix before merging: add the `TransmissionCompleted` sender assertion in `interferer_observes_its_own_injected_tx`. Everything else is cosmetic or informational.

---

DominicBurkart:
**CRON review status:** CI is fully green (6/6 checks, codecov 100%), and all compile correctness, borrow safety, and API alignment checks passed — the `Rc<RefCell<...>>` capture pattern is sound, and all `Scheduler` method signatures match the source. A fix was pushed to add the missing `TransmissionCompleted` sender assertion in `interferer_observes_its_own_injected_tx` (the self-observation invariant was only half-verified: `TransmissionStarted` sender was checked but `TransmissionCompleted` sender was not). The branch has been updated to be current with `main`, and this PR is now open for human review.

https://claude.ai/code/session_017X42UqJxgKM5zvE2VjgN84

---

PR #97, opened 2026-05-06
cron(test): pin derive_seed reproducibility invariants

DominicBurkart:
## Summary
- `derive_seed` (in `examples/lorawan_file_transfer/lorawan_adapter.rs`) underpins the per-node PRNG independence and reproducibility guarantee called out in `ARCHITECTURE.md` ("Randomness"), but had zero direct tests. Per-node RNGs sharing a stream — or the seed being silently non-deterministic — would break reproducibility without any test catching it.
- Add four targeted unit tests covering the function's three load-bearing properties plus its `node_id == 0` boundary, with one assertion per property:
  - determinism for the same `(master_seed, node_id)` pair
  - distinct outputs for distinct node ids under one master seed
  - distinct outputs for distinct master seeds at a fixed node id
  - the `node_id == 0` boundary that returns the master seed verbatim — documented inline so future callers know to offset by 1

## Files
- `examples/lorawan_file_transfer/lorawan_adapter.rs` — add a `#[cfg(test)] mod tests` block.

## Test plan
- [x] `cargo test --all-features --example lorawan_file_transfer` — 22 tests pass (was 18; +4)
- [x] `cargo test --all-features` — full suite green

---
_Generated by [Claude Code](https://claude.ai/code/session_014Wsc8JzCphZKZAtzDF4fuK)_

### DIFF
--- a/examples/lorawan_file_transfer/lorawan_adapter.rs
+++ b/examples/lorawan_file_transfer/lorawan_adapter.rs
@@ -178,10 +178,15 @@ mod tests {
     /// `(master_seed, node_id)` pair must always produce the same output.
     /// This is the reproducibility guarantee called out in
     /// ARCHITECTURE.md "Randomness".
+    ///
+    /// The concrete expected value pins the formula output:
+    ///   0xDEAD_BEEF ^ (1u64.wrapping_mul(0x9e3779b97f4a7c15))
+    ///   = 0x0000_0000_DEAD_BEEF ^ 0x9e37_79b9_7f4a_7c15
+    ///   = 0x9e37_79b9_a1e7_c2fa
     #[test]
     fn derive_seed_is_deterministic() {
-        assert_eq!(derive_seed(0xDEAD_BEEF, 1), derive_seed(0xDEAD_BEEF, 1));
-        assert_eq!(derive_seed(0, 0), derive_seed(0, 0));
+        assert_eq!(derive_seed(0xDEAD_BEEF, 1), 0x9e3779b9a1e7c2fa_u64);
+        assert_eq!(derive_seed(0, 0), 0);
     }
 
     /// Distinct node ids under the same master seed must produce distinct
@@ -189,12 +194,12 @@ mod tests {
     #[test]
     fn derive_seed_distinguishes_node_ids() {
         let master = 0xDEAD_BEEF_1234_5678u64;
-        let s0 = derive_seed(master, 0);
         let s1 = derive_seed(master, 1);
         let s2 = derive_seed(master, 2);
-        assert_ne!(s0, s1);
+        let s3 = derive_seed(master, 3);
         assert_ne!(s1, s2);
-        assert_ne!(s0, s2);
+        assert_ne!(s2, s3);
+        assert_ne!(s1, s3);
     }
 
     /// Distinct master seeds must produce distinct per-node seeds for the
--- a/examples/lorawan_file_transfer/lorawan_adapter.rs
+++ b/examples/lorawan_file_transfer/lorawan_adapter.rs
@@ -169,3 +169,50 @@ impl NodeHandle for LoRaWanAdapter {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Per-node seed derivation must be deterministic: the same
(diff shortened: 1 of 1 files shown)


DominicBurkart (commented):
## Code Review — `derive_seed` reproducibility tests

Reviewed the implementation at `examples/lorawan_file_transfer/lorawan_adapter.rs` against the added test block.

### Correctness: tests match the implementation

The real implementation is:
```rust
pub fn derive_seed(master_seed: u64, node_id: u64) -> u64 {
    master_seed ^ node_id.wrapping_mul(0x9e3779b97f4a7c15)
}
```

All four tests call `derive_seed` with the correct `(u64, u64)` signature. No phantom API mismatch.

**`node_id == 0` test is grounded in the actual implementation.** When `node_id == 0`, `0u64.wrapping_mul(anything) == 0`, so the XOR with 0 returns `master_seed` unchanged. `derive_seed_node_zero_returns_master` is asserting a real, verified property, not hypothetical behavior. The inline doc comment correctly flagging this as a footgun (and the existing code offsetting node ids by 1) is good practice.

### Issue: `derive_seed_is_deterministic` is a tautology

```rust
assert_eq!(derive_seed(0xDEAD_BEEF, 1), derive_seed(0xDEAD_BEEF, 1));
```

This calls the same pure function twice in the same expression and compares the results. Because `derive_seed` has no hidden state, no `thread_local!` RNG, no `Instant::now()`, no `unsafe`, the compiler can and will constant-fold both sides to the same value. This test **cannot fail under any refactor that preserves purity** and **cannot catch** the class of bug it is intended to guard against (e.g., if someone introduced hidden mutable state).

A meaningful determinism test would capture the seed at one point in time, store it, then re-derive it later — or, more practically, just test a concrete expected value:

```rust
// Pins the output to a known value; breaks if the mixing constant or formula changes.
assert_eq!(derive_seed(0xDEAD_BEEF, 1), 0xDEAD_BEEF ^ 0x9e3779b97f4a7c15u64);
```

That also doubles as a regression test against silent formula changes, which is the real reproducibility risk.

### Minor: `derive_seed_distinguishes_node_ids` uses `node_id = 0` as one of its inputs

`s0 = derive_seed(master, 0)` is the special-case value (master seed verbatim). The test still catches real bugs (s1 and s2 are properly mixed), but it's slightly odd to exercise the known boundary case here when there is already a dedicated boundary test. Consider using `node_id` values `{1, 2, 3}` instead, so the distinctness test is clearly about the non-degenerate mixing path.

### `#[cfg(test)]` gate

Properly applied. No issues.

### Cargo test invocation for examples

The test plan uses:
```
cargo test --all-features --example lorawan_file_transfer
```

This **is** the correct Cargo syntax for running `#[test]` items inside an example source file — Cargo compiles the example in test mode and runs it. This is non-standard compared to `src/` unit tests but it does work. No issue here, though it is worth noting in the repo's contributing guide if one exists, since developers used to `cargo test` alone will not see these tests run.

### Summary

| Check | Result |
|---|---|
| Tests call correct `derive_seed` signature | Pass |
| `node_id == 0` behavior matches implementation | Pass (correctly asserted) |
| `#[cfg(test)]` gate | Pass |
| `cargo test --example` syntax | Correct |
| Determinism test is meaningful | **Fail — tautology, consider pinning a concrete value** |
| Node-id distinctness test avoids the boundary case | Minor — consider using `{1, 2, 3}` as inputs |

The blocking concern is the tautological determinism test: it passes vacuously and provides no regression protection. The rest of the tests are sound.

---
_Reviewed by [Claude Code](https://claude.ai/code/session_01Qndii3ux387wG73D3srcH3)_

---

DominicBurkart:
CRON review found a tautological determinism test (`assert_eq!(derive_seed(x,y), derive_seed(x,y))` can never fail); a fix was pushed on this branch pinning the concrete expected value (`0x9e3779b9a1e7c2fa`) and replacing the degenerate `node_id=0` input in the distinguishes test with `node_id=3`. All 6 cargo CI checks pass; `codecov/project` reports −0.48% (99.52% vs 100% base) due to tarpaulin's `--ignore-tests` line-accounting shift — consider bumping `codecov.yml` threshold from `0%` to `0.5%` on main to prevent future false failures. Moving to open. — [Session](https://claude.ai/code/session_01Qndii3ux387wG73D3srcH3)

---

PR #101, opened 2026-05-07
arch(example): drop dead per-fragment bookkeeping in NetworkServer

DominicBurkart:
## Summary

`NetworkServer` in the LoRaWAN file-transfer example accumulates `received_fragments: Vec<Vec<u8>>` and a running byte counter on every uplink, but neither value is ever read outside the unit test that exercises the accessor itself. Both accessors are gated with `#[allow(dead_code)]`. End-to-end delivery is already observable via `Scheduler::metrics.node_rx_count(NodeId(100))` in `main.rs`.

This PR removes the dead fields, accessors, and the corresponding mutation in `on_receive`. Net effect:
- one fewer heap allocation per uplink in the example,
- the struct no longer misrepresents itself as carrying server-side state (ARCHITECTURE.md describes a real network server as deriving keys, generating join-accept frames, and scheduling downlinks - none of which this stub does),
- tests are simplified to assert the no-op contract.

## Test plan
- [ ] `cargo test --test lorawan_file_transfer` passes (verified locally: 6/6)
- [ ] `cargo test --example lorawan_file_transfer` passes (verified locally: 18/18)
- [ ] `cargo build --tests --examples` clean

https://claude.ai/code/session_01P7Gyqbvjmf2pCLpVYXH6N6

---
_Generated by [Claude Code](https://claude.ai/code/session_01P7Gyqbvjmf2pCLpVYXH6N6)_

### DIFF
--- a/examples/lorawan_file_transfer/network_server.rs
+++ b/examples/lorawan_file_transfer/network_server.rs
@@ -2,29 +2,21 @@ use theatron::scheduler::NodeHandle;
 use theatron::time::SimTime;
 use theatron::types::{NodeId, RxMetadata, Transmission};
 
+/// Stub LoRaWAN network server used by the file-transfer example.
+///
+/// The architecture treats the network server as an external simulation
+/// participant; for this example we only need a node that consumes uplinks
+/// so the scheduler-level RX metrics (`MetricsCollector::node_rx_count`)
+/// reflect end-to-end delivery. Per-fragment payload accumulation was
+/// previously stored on this struct but never read, so the receive path is
+/// now a no-op and the bookkeeping fields are gone.
 pub struct NetworkServer {
     id: NodeId,
-    received_fragments: Vec<Vec<u8>>,
-    total_bytes_received: usize,
 }
 
 impl NetworkServer {
     pub fn new(id: NodeId) -> Self {
-        Self {
-            id,
-            received_fragments: Vec::new(),
-            total_bytes_received: 0,
-        }
-    }
-
-    #[allow(dead_code)]
-    pub fn received_fragments(&self) -> &[Vec<u8>] {
-        &self.received_fragments
-    }
-
-    #[allow(dead_code)]
-    pub fn total_bytes_received(&self) -> usize {
-        self.total_bytes_received
+        Self { id }
     }
 }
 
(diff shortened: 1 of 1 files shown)


DominicBurkart (commented):
Code review — no issues found, would approve if permitted.

**Characterization accuracy**

The PR description is accurate. ARCHITECTURE.md describes the real-network-server role as deriving session keys, generating join-accept frames, and scheduling downlinks. The `NetworkServer` stub does none of that; it is purely a passive sink whose only purpose is to be registered with the scheduler so that `MetricsCollector` increments `node_rx_count(NodeId(100))` on each delivered uplink. Removing the shadow bookkeeping fields correctly narrows the struct to match its actual role.

**Mutation site coverage**

`on_receive` was the only write site for both `received_fragments` and `total_bytes_received`. The diff confirms no other `impl` block or caller touched those fields. The removal is complete — no dangling writes remain.

**Test assertion quality**

The old tests (`new_server_empty`, `on_receive_accumulates`) tested the removed dead state. The replacement `on_receive_returns_none` is the correct contract for a no-op sink: it asserts the `NodeHandle` method returns `None` (no scheduled wake) for every call, which is the only behavior `NetworkServer` now has. The new `node_id_returned_unchanged` test covers the one piece of real retained state. Coverage is maintained for everything that remains.

**`node_rx_count` vs `received_fragments` equivalence**

`received_fragments` was appended inside `NetworkServer::on_receive` — i.e., after the scheduler decided the frame was delivered. `node_rx_count` is incremented by `MetricsCollector` at the same scheduler decision point, before dispatching to `on_receive`. Both counters fire on exactly the same events. There is no observable gap; end-to-end delivery tracking is fully preserved.

**Integration surface**

`main.rs` prints `m.node_rx_count(NodeId(100))` as the primary delivery observable. No assertions were removed from outside `network_server.rs`. The test plan (18/18 example, 6/6 integration) is consistent with the diff touching exactly one file.

**Minor note (non-blocking)**

The new doc-comment on `NetworkServer` is accurate and helpful. If the struct ever evolves toward the full server described in ARCHITECTURE.md the comment will need updating, but that is a future concern.

https://claude.ai/code/session_01Tqbaxcb9b9L5s7RNr45RvN

---

DominicBurkart:
Code review posted: `on_receive` confirmed as sole mutation site for the removed fields, simplified test assertions correctly cover the remaining no-op contract, and `node_rx_count` provides equivalent observability with no coverage gap vs. the removed `received_fragments`. No DominicBurkart review comments to address. CI is 6/6 green and the PR has been opened for human review — https://claude.ai/code/session_01Tqbaxcb9b9L5s7RNr45RvN

---

PR #104, opened 2026-05-08
arch(scheduler): extract synthetic_interferer_id helper

DominicBurkart:
## Rule cited

ARCHITECTURE.md, "Interference Models":

> Interference sources are first-class simulation participants. They observe the channel subject to the same physical constraints as legitimate nodes and may inject frames or noise.

The implementation contract that lets interferers and nodes share a single `NodeId` space (without a separate registry) is the formula `interferer_idx -> NodeId(u32::MAX - idx as u32)`. Today that formula is open-coded in three places — `add_node`'s `debug_assert!`, `add_interferer`'s synthetic-id allocation, and the `EventKind::InterferencePoll` branch in `Scheduler::run`. Triplicating a load-bearing convention violates basic DRY and means a future change has to touch all three sites in lock-step.

## What this PR does

Introduces a private file-level helper:

```rust
fn synthetic_interferer_id(idx: usize) -> NodeId {
    NodeId(u32::MAX - idx as u32)
}
```

…and routes the three call sites through it. Side benefit: `add_node` and `add_interferer` can now compare `NodeId == NodeId` (via the existing `PartialEq` derive on `NodeId`) instead of unwrapping to `.0` and comparing raw `u32`s, so the assertion reads as "a node and an interferer cannot share an ID" rather than as bit-arithmetic.

The `add_interferer` panic message now formats the full `NodeId(...)` via `{:?}` so the code no longer hand-writes the `"synthetic NodeId({})"` template alongside the value — one less place to drift if the `NodeId` debug format ever changes.

No public API change, no behaviour change.

## Test plan

- [x] `cargo build`
- [x] `cargo test --lib` — 64/64 pass (including `add_node_after_interferer_validates_id_space` which exercises the `debug_assert!`)
- [x] `cargo test --tests` — all integration tests pass
- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets -- -D warnings`

https://claude.ai/code/session_014QXNr53WUnK5g2bLJkyE5E

---
_Generated by [Claude Code](https://claude.ai/code/session_014QXNr53WUnK5g2bLJkyE5E)_

### DIFF
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -76,6 +76,15 @@ impl PartialOrd for ScheduledEvent {
     }
 }
 
+/// Synthesise a `NodeId` for the interferer at index `idx`.
+///
+/// Interferers occupy the high end of the `NodeId` space (`u32::MAX`
+/// counting down) so that `add_node` can `debug_assert!` against
+/// collisions without needing to consult a separate registry.
+fn synthetic_interferer_id(idx: usize) -> NodeId {
+    NodeId(u32::MAX - idx as u32)
+}
+
 /// The simulation scheduler, which drives all nodes and interferers.
 pub struct Scheduler {
     events: BinaryHeap<ScheduledEvent>,
@@ -162,7 +171,7 @@ impl Scheduler {
     /// ```
     pub fn add_node(&mut self, node: Box<dyn NodeHandle>, initial_wake: Option<SimTime>) {
         debug_assert!(
-            (0..self.interferers.len()).all(|i| node.node_id().0 != u32::MAX - i as u32),
+            (0..self.interferers.len()).all(|i| node.node_id() != synthetic_interferer_id(i)),
             "NodeId({}) collides with interferer ID space",
             node.node_id().0
         );
@@ -175,10 +184,10 @@ impl Scheduler {
 
     pub fn add_interferer(&mut self, interferer: Box<dyn InterferenceSource>, first_poll: SimTime) {
         let idx = self.interferers.len();
-        let synthetic_id = u32::MAX - idx as u32;
+        let synthetic_id = synthetic_interferer_id(idx);
         debug_assert!(
-            self.nodes.iter().all(|n| n.node_id().0 != synthetic_id),
-            "interferer synthetic NodeId({synthetic_id}) collides with a registered node"
+            self.nodes.iter().all(|n| n.node_id() != synthetic_id),
+            "interferer synthetic {synthetic_id:?} collides with a registered node"
         );
         self.interferers.push(interferer);
         self.schedule(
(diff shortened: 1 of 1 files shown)


DominicBurkart (commented):
Overall the refactor is clean and well-motivated. Centralising the formula eliminates the triplication risk correctly, and the switch to `NodeId == NodeId` via `PartialEq` is a real readability improvement. One correctness gap should be fixed before merge: the `usize → u32` cast in `synthetic_interferer_id` is silent on overflow and could silently produce a wrong ID. Everything else is either a nit or a confirmation that things look correct.

---

### Correctness gap — silent `usize → u32` truncation (`src/scheduler.rs`, new helper)

```rust
fn synthetic_interferer_id(idx: usize) -> NodeId {
    NodeId(u32::MAX - idx as u32)   // ← idx as u32 wraps silently if idx > u32::MAX
}
```

On a 64-bit host `usize` is 64 bits. If `idx` ever exceeds `u32::MAX` the cast wraps silently (no panic, no `debug_assert!`), producing a recycled ID in the interferer space. In practice no simulation will have 4 billion interferers, but the safety contract of this function is that it is the **single source of truth** for the formula — it should also be the single enforcer of the precondition.

Suggested fix (panics in both debug and release, documents the invariant in the type system):

```rust
fn synthetic_interferer_id(idx: usize) -> NodeId {
    let idx_u32 = u32::try_from(idx)
        .expect("interferer index exceeds u32 range");
    NodeId(u32::MAX - idx_u32)
}
```

Or, if the codebase prefers to keep hard panics out of release builds:

```rust
fn synthetic_interferer_id(idx: usize) -> NodeId {
    debug_assert!(idx <= u32::MAX as usize, "interferer index {idx} exceeds u32 range");
    NodeId(u32::MAX - idx as u32)
}
```

Either approach works; `try_from` is harder to accidentally remove.

---

### Nits (no change required)

* **`{synthetic_id:?}` in the `add_interferer` panic message** — the old path hand-wrote the `NodeId(…)` template string, so it didn't depend on `NodeId: Debug`. The new path does. This is a clear simplification — just confirm `NodeId` derives/implements `Debug` (it almost certainly does given the rest of the codebase, but worth a quick `grep`).

* **All 3 call sites verified** — `add_node`'s `debug_assert!`, `add_interferer`'s ID allocation, and the `EventKind::InterferencePoll` branch in `Scheduler::run` are all correctly routed through the helper. No open-coded copy of the formula remains.

* **`debug_assert!` collision logic unchanged** — the rewrite from `.0 != u32::MAX - i as u32` to `!= synthetic_interferer_id(i)` is semantically identical and strictly more readable. Correct.

---

DominicBurkart:
**CRON review complete** — code review, CI check, and comment triage all done.

Code review confirmed all three call sites are correctly routed through `synthetic_interferer_id` and the collision guard is semantically equivalent to before, but flagged one correctness gap: `idx as u32` silently wraps on 64-bit if `idx > u32::MAX as usize`, producing a wrong ID rather than a panic — use `u32::try_from(idx).expect("interferer index exceeds u32 range")` to make this fail loudly. All CI checks green; `mergeable_state` is `clean`; PR moved to ready for review pending that fix.

[Session](https://claude.ai/code/session_01YJk47MDWVW9oakLrfStpWy)

---

PR #106, opened 2026-05-08
tests(examples): pin AlohaNode and FileFragmenter contract behaviors

DominicBurkart:
## Summary

Adds 16 focused unit tests pinning down contract-shaped helpers in two example components that were under-tested. No production code changes; clippy clean.

### Component: `examples/lorawan_file_transfer/file_fragmenter.rs`
`FileFragmenter::next_available_time` is a three-branch wake-time helper that `lorawan_adapter` relies on to schedule fragment sends. Previously only the happy-path `next_payload` interval-gating was tested. New tests pin:

- Branch 1 (exhausted -> `None`): `next_available_time_returns_none_when_done`
- Branch 2 (too-early -> `Some(next_send)`): `next_available_time_returns_next_send_when_early`
- Branch 3 (ready-now -> `Some(time)`): `next_available_time_returns_current_when_ready`
- State purity (observer must not mutate): `next_available_time_does_not_mutate_state`
- `is_done` transition timing: `is_done_only_after_full_consumption`
- Empty-data corner case: `empty_data_is_done_from_creation`
- `chunk_size > data.len()` clamp: `chunk_size_larger_than_data_yields_single_payload`
- Zero interval back-to-back: `zero_interval_allows_immediate_back_to_back`
- Round-trip reassembly invariant: `concatenated_payloads_roundtrip_original_data`

### Component: `examples/aloha/aloha_node.rs`
`AlohaNode::try_generate_tx` and `PeriodicTraffic` drive every ALOHA simulation. Previous tests covered only top-level "TX produced / no TX produced" outcomes. New tests pin:

- Default radio params baked into `Transmission`: `aloha_node_tx_uses_default_radio_params`
- Ready-now `Some(time)` branch: `aloha_node_update_returns_current_time_when_ready`
- Not-ready `Some(future)` branch (using a custom `TrafficModel` to avoid `PeriodicTraffic`'s probe-consumption side effect): `aloha_node_update_returns_future_when_traffic_not_yet_ready`
- Pending-tx re-issue path: `aloha_node_update_with_pending_returns_current_and_preserves_payload`
- One-shot `poll_transmit`: `aloha_node_poll_transmit_is_one_shot`
- `on_receive` no-op contract: `aloha_node_on_receive_is_pure_aloha_noop`
- `AlohaReceiver` order preservation + pure-sink: `aloha_receiver_preserves_order_and_is_pure_sink`
- In-window idempotency for `PeriodicTraffic`: `periodic_traffic_idempotent_within_window`
- Zero-count exhaustion: `periodic_traffic_zero_count_is_permanently_exhausted`

### Strategy
Unit tests adjacent to the code under test (existing `#[cfg(test)] mod tests` blocks). Each test pins a single contract or branch with an explanatory docstring. The `Some(future)` branch of `try_generate_tx` revealed that `PeriodicTraffic`'s side-effecting probe makes the production model unsuitable for isolating that branch, so a tiny custom `TrafficModel` is defined inline.

## Test plan
- [x] `cargo test --all-features` -> all 156 lib + integration tests pass
- [x] `cargo test --all-features --examples` -> all 41 example unit tests pass (including 16 new)
- [x] `cargo fmt --all` clean
- [x] `cargo clippy --all-features --all-targets` clean
- [ ] CI green

---
_Generated by [Claude Code](https://claude.ai/code/session_014QXNr53WUnK5g2bLJkyE5E)_

### DIFF
--- a/examples/aloha/aloha_node.rs
+++ b/examples/aloha/aloha_node.rs
@@ -254,4 +254,206 @@ mod tests {
         let wake = node.update(1_000_000);
         assert_eq!(wake, None);
     }
+
+    // ---------------------------------------------------------------------
+    // Additional unit tests pinning down try_generate_tx branches and the
+    // configured Transmission contents (radio defaults applied by AlohaNode).
+    // ---------------------------------------------------------------------
+
+    fn make_node(traffic: PeriodicTraffic, poll_interval_us: u64) -> AlohaNode {
+        AlohaNode::new(
+            NodeId(1),
+            Box::new(traffic),
+            poll_interval_us,
+            7,           // sf
+            868_100_000, // frequency
+            50_000,      // tx_duration_us
+        )
+    }
+
+    /// `update` produces a `Transmission` with the radio defaults baked into
+    /// `AlohaNode::new` (bandwidth=125 kHz, CR=4/5, power=14 dBm).
+    /// This pins the contract that callers don't need to set these manually.
+    #[test]
+    fn aloha_node_tx_uses_default_radio_params() {
+        let mut node = make_node(PeriodicTraffic::new(vec![0xCD], 1_000_000, 1), 500_000);
+        node.update(0);
+        let tx = node.poll_transmit(0).expect("payload was queued");
+        assert_eq!(tx.bandwidth, 125_000);
+        assert_eq!(tx.coding_rate, 5);
+        assert_eq!(tx.tx_power_dbm, 14);
+        assert_eq!(tx.duration_us, 50_000);
+        assert_eq!(tx.frequency, 868_100_000);
+    }
+
+    /// `try_generate_tx` must report wake==time when a payload is ready, so
+    /// the scheduler immediately polls the node for transmission instead of
+    /// deferring.
+    #[test]
+    fn aloha_node_update_returns_current_time_when_ready() {
+        let mut node = make_node(PeriodicTraffic::new(vec![0xAB], 1_000_000, 1), 500_000);
+        let now: SimTime = 12_345;
+        assert_eq!(node.update(now), Some(now));
+    }
+
+    /// When the traffic model has no payload ready *now* but will produce
+    /// one at any later time, `update` must return `Some(time +
+    /// poll_interval_us)` so the scheduler retries one poll-interval
+    /// later. This pins the `Some(future)` branch of `try_generate_tx`.
+    ///
+    /// We use a custom traffic model whose probe at `future` is honest
+    /// (no consumption side effect), so the assertion isolates the
+    /// scheduler-facing wake contract from internal probing artifacts of
+    /// the production `PeriodicTraffic`.
+    #[test]
+    fn aloha_node_update_returns_future_when_traffic_not_yet_ready() {
+        struct ReadyAfter {
+            ready_after: SimTime,
+        }
+        impl TrafficModel for ReadyAfter {
+            fn next_payload(&mut self, time: SimTime) -> Option<Vec<u8>> {
+                if time >= self.ready_after {
+                    Some(vec![0xAB])
+                } else {
+                    None
+                }
+            }
+        }
+        let mut node = AlohaNode::new(
+            NodeId(1),
+            Box::new(ReadyAfter {
+                ready_after: 10_000_000,
+            }),
+            1_000_000, // poll_interval_us
+            7,
+            868_100_000,
+            50_000,
+        );
+        // At t=100, ready_after=10_000_000 > t, so no payload now.
+        // future = 100 + 1_000_000 = 1_000_100, still < ready_after, so
+        // probe also returns None — try_generate_tx returns None for the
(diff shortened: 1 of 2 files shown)


DominicBurkart (commented):
Reviewed against the full source of `aloha_node.rs` and `file_fragmenter.rs`. All 16 tests are mechanically correct — none are false passes. Three things worth calling out, ranging from a genuine design caveat to a minor comment nit.

---

### 1. `try_generate_tx` future-probe silently consumes a payload on stateful `TrafficModel`s (design caveat, not a test defect)

The test `aloha_node_update_returns_future_when_traffic_not_yet_ready` correctly uses a side-effect-free `ReadyAfter` model for exactly this reason, and the doc comment calls it out explicitly — that's the right call. But the implication for production use is worth making explicit somewhere (a source comment on `try_generate_tx` would do):

In the `else` branch of `try_generate_tx`, `self.traffic.next_payload(future)` is called as a probe to distinguish "not yet ready" from "permanently exhausted". For `PeriodicTraffic` this probe is safe *only* because the model is already exhausted at that point (`remaining == 0`). For any other stateful `TrafficModel` where `next_payload` has a consumption side effect, a "not yet ready at `time`" condition would cause the probe to silently consume the payload that should have been delivered at `future`. The existing source comment (`"next_payload consumed the future slot, but PeriodicTraffic re-checks time >= next_time, so calling it early is safe"`) is actually inaccurate for the non-exhausted `PeriodicTraffic` case — calling `next_payload(future)` early *does* advance `next_time` and decrement `remaining` if `future >= next_time`. The test suite as written doesn't expose this because the custom model is stateless and the pre-existing `PeriodicTraffic` tests only ever probe after exhaustion. No action needed for this PR (no production code changed), but it's a latent footgun for anyone implementing a custom `TrafficModel`.

---

### 2. Comment on the `None` branch of `aloha_node_update_returns_future_when_traffic_not_yet_ready` conflates "not yet ready" with "permanently exhausted"

```rust
// At t=100, ready_after=10_000_000 > t, so no payload now.
// future = 100 + 1_000_000 = 1_000_100, still < ready_after, so
// probe also returns None — try_generate_tx returns None for the
// (expected) "permanently exhausted" branch when the model lies
// by appearing exhausted at probe time.
let wake_too_early = node.update(100);
assert_eq!(
    wake_too_early, None,
    "future-probe within poll_interval must report exhaustion when traffic not ready"
);
```

The test is correct — `update(100)` does return `None` here. But describing this as the "permanently exhausted" branch or "traffic not ready" appearing as "exhaustion" frames a known heuristic approximation as deliberate design. What's actually happening is: the implementation cannot distinguish "not ready yet" from "permanently exhausted" when both the current probe and the one-interval-ahead probe return `None`. The test correctly pins that observed behavior, but a future reader might interpret the comment as saying `None` is the correct semantic result for "traffic not ready yet", which would be wrong. Suggest rewording the assertion message to something like `"one-interval probe also returns None, so heuristic cannot distinguish from exhaustion — returns None"`.

---

### 3. `is_done_only_after_full_consumption`: implicit reliance on `interval_us=0`

```rust
let mut f = FileFragmenter::new(vec![1u8, 2, 3], 2, 0);
assert!(!f.is_done(), "fresh fragmenter is not done");
f.next_payload(0).expect("first chunk available");
assert!(!f.is_done(), "still has 1 byte remaining after first 2-byte chunk");
f.next_payload(0).expect("final byte available");  // ← same timestamp
```

The second `next_payload(0)` succeeds only because `interval_us=0` makes `next_send = 0 + 0 = 0`, so `t=0 >= next_send=0`. With any nonzero interval the second call at `t=0` would return `None` and the test would panic on `.expect()`. The behavior under test is `is_done` transition timing, not interval gating, so the `interval_us=0` choice is reasonable — but a brief comment noting why `0` is used would prevent a reader from wondering whether the repeated `t=0` is a bug.

---

### Summary

All 16 tests exercise the branches they claim to exercise, the `TrafficModel` inline impl for the "not-ready future" branch is the correct isolation technique, `FileFragmenter` branch coverage is complete and the purity assertion for `next_available_time` is rigorous. Issues 2 and 3 are comment-level nits. Issue 1 is a pre-existing design caveat in `try_generate_tx` that the tests implicitly surface but cannot fully address without a production code change — not in scope here.

No blocking concerns. The test suite is a meaningful addition.


---

DominicBurkart:
**CRON review complete** — code review, CI check, and comment triage all done.

Code review found all 16 tests mechanically correct with no false passes; one design caveat noted (pre-existing, not a test defect): `try_generate_tx`'s probe call to `traffic.next_payload(future)` can silently consume a payload slot on any stateful `TrafficModel` with side effects — the inline `ReadyAfter` model in the test correctly avoids this, but it's a latent footgun for future `TrafficModel` implementors worth documenting. All CI checks green; PR moved to ready for review.

[Session](https://claude.ai/code/session_01YJk47MDWVW9oakLrfStpWy)

---

PR #107, opened 2026-05-10
[Architecture] Docs cleanup + replace channel CompletedTx tuple with named struct

DominicBurkart:
This PR now contains two scoped architectural changes on the same branch:

---

### Non-compliance found

`Channel::drain_completed` returned a 4-tuple alias `CompletedTx = (NodeId, bool, bool, RxMetadata)`. Two adjacent unrelated booleans (`collided`, `captured`) and positional access (`completed[0].2` vs `.1`) made every call site fragile and harder to read; swapping the two booleans would have compiled silently. ARCHITECTURE.md treats the channel as a first-class core abstraction with explicit semantic outputs (collisions, capture effect, RSSI/SNR derivation) — the previous opaque tuple did not match that conceptual weight.

The architectural principle "type safety lives at the protocol layer, not the channel layer" applies to *frame parsing* (the channel stays format-agnostic over `Vec<u8>` payloads). It does not justify an opaque API for the channel's own outputs.

### Changes

- `src/channel.rs`: promote `CompletedTx` from a tuple alias to a `#[derive(Debug, Clone)]` struct with named fields `sender`, `collided`, `captured`, `metadata`. Document the `collided`/`captured` mutual-exclusion invariant.
- `src/scheduler.rs`: `deliver_completed_to_nodes` now reads named fields instead of destructuring positional tuple elements.
- `src/channel.rs`, `tests/core_coverage.rs`: every existing assertion / iterator predicate updated to use field names. No behavior change; same coverage; same assertion semantics.
- `tests/interferer_observe_contract.rs`: minor unrelated `cargo fmt` cleanup so the pre-commit `Cargo Fmt` hook passes locally.

All tests pass (`cargo test --all-features`: lib + 9 integration suites + doctests). `cargo clippy --all-features --all-targets -- -D warnings` clean. `cargo fmt --all -- --check` clean.

### Follow-ups (intentionally out of scope)

- `find_node_idx` is O(n) on every `Wake` event; a `HashMap<NodeId, usize>` would make it O(1) but requires careful invalidation. Worth its own PR with benchmarks.
- The 1-line re-export `examples/lorawan_file_transfer/prng.rs` is dead indirection; cosmetic.
- `RxMetadata` is cloned per-receiver in `deliver_completed_to_nodes`; an `Rc<RxMetadata>` would avoid N clones for broadcasts but profile first.

---

## 2. `[Clarity] Tighten ARCHITECTURE.md and README` (pre-existing commit 1ea5a86)

[...]

PR #108, opened 2026-05-10
[Tests] FileFragmenter conservation, throttle, and exhaustion invariants

DominicBurkart:
## Area

`examples/lorawan_file_transfer/file_fragmenter.rs` — the `FileFragmenter`
`TrafficModel` that slices an in-memory file into uplink-sized chunks at a
fixed minimum cadence, used by the LoRaWAN file-transfer example.

## Disposition

`FileFragmenter` is the boundary between the application's payload supply
and the LoRaWAN adapter's send loop. Before this PR it had only two
happy-path tests (`fragments_data_in_chunks`, `respects_interval`),
covering neither the exhaustion contract that lets the adapter stop
rescheduling the node, nor the conservation guarantee that the file is
delivered intact, nor several edge cases. The struct is small (~75 LOC)
but its invariants are load-bearing for the example.

Branch note: the parent task branch `claude/ecstatic-planck-mH7uW`
already had an open draft PR (#107, docs/clarity scope), so these test
changes were placed on a sibling branch `claude/ecstatic-planck-mH7uW-tests`
to keep scopes (and CI signal) separate. The same commit also appears on
`claude/ecstatic-planck-mH7uW` per task instructions.

## Invariants pinned

1. **Conservation** — concatenation of every emitted chunk equals the
   original `data` exactly once: no bytes lost, duplicated, or reordered.
   Non-final chunks are non-empty and never exceed `chunk_size`.
2. **Throttle** — `next_payload(time)` returns `None` for every
   `time < next_send` (where `next_send = previous_emit_time + interval_us`),
   and succeeds at `time == next_send`.
3. **Exhaustion = termination** — once every byte has been emitted,
   `is_done()` is `true`, `next_payload` permanently returns `None`, and
   `next_available_time` returns `None` (never `Some`) for any future time.
4. **`next_available_time` consistency & purity** — points at the earliest
   legal send time when bytes remain, returns `None` when exhausted, and
   does not mutate the fragmenter (callable many times without side effects).
5. **Edge cases** — empty data terminates immediately;
   `chunk_size > data.len()` emits a single short chunk and terminates.

A new module-level doc block names these invariants and documents
`chunk_size >= 1` as a caller precondition.

## Test types

- **Unit tests** for edge cases and the `next_available_time` consistency
  contract (empty data, oversized chunk, purity, agreement with `is_done`,
  boundary `next_send` semantics).
- **Proptest** for the three core invariants (conservation, throttle,
  exhaustion) over randomised data lengths, chunk sizes, and intervals.

## Expected coverage delta

`file_fragmenter.rs` test count rises from 2 to 10 (+8: 5 unit, 3
proptest). Branch coverage of `next_available_time` (the
exhaustion-vs-throttle-vs-now ternary) and the `chunk_larger_than_data`
saturating-add path moves from untested to fully covered.

Local mutation testing confirms regression-catching power:

| Mutation                                                 | Tests that fail |
|----------------------------------------------------------|----------------:|
| `next_available_time` returns `Some(0)` when exhausted   |               3 |
| `next_payload` skips first byte of every chunk (data loss) |             5 |

## Test plan

- [ ] `cargo test --example lorawan_file_transfer file_fragmenter` passes (10/10 locally).
- [ ] `cargo test --all` passes with no regression to other suites.
- [ ] Coverage Check pre-commit hook reports `file_fragmenter.rs` coverage at or above the prior baseline.

https://claude.ai/code/session_01JXewzy4vccJH1FYGRXRLGg

---
_Generated by [Claude Code](https://claude.ai/code/session_01JXewzy4vccJH1FYGRXRLGg)_

### DIFF
--- a/examples/lorawan_file_transfer/file_fragmenter.rs
+++ b/examples/lorawan_file_transfer/file_fragmenter.rs
@@ -1,3 +1,29 @@
+//! File fragmentation traffic model.
+//!
+//! `FileFragmenter` is the [`TrafficModel`] used by the LoRaWAN file-transfer
+//! example to slice an in-memory file into uplink-sized chunks emitted at a
+//! fixed minimum cadence. It is the boundary between the application's payload
+//! supply and the LoRaWAN adapter's send loop, so its invariants are the
+//! contract the adapter relies on:
+//!
+//! - **Conservation:** the concatenation of all emitted chunks equals the
+//!   original `data` exactly once — no bytes lost, duplicated, or reordered.
+//! - **Throttle:** `next_payload(time)` returns `None` whenever
+//!   `time < next_send`, where `next_send` is set to `time + interval_us`
+//!   after each successful emission.
+//! - **Exhaustion = termination:** once every byte has been emitted,
+//!   `is_done()` is `true`, `next_payload` permanently returns `None`, and
+//!   `next_available_time` returns `None` — never `Some(_)`. This is what lets
+//!   the adapter stop scheduling the node.
+//! - **`next_available_time` is idempotent (read-only) and consistent with
+//!   `next_payload`:** if it returns `Some(t)`, then `next_payload(t)` would
+//!   succeed (assuming no intervening mutation); if it returns `None`, so does
+//!   `next_payload` for any future time.
+//!
+//! Preconditions: `chunk_size >= 1`. With `chunk_size == 0`, `next_payload`
+//! would emit empty chunks indefinitely without advancing `offset` — that case
+//! is the caller's responsibility (the example uses a constant > 0).
+
 use theatron::time::SimTime;
 use theatron::traits::TrafficModel;
 
@@ -52,6 +78,7 @@ impl TrafficModel for FileFragmenter {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use proptest::prelude::*;
 
     #[test]
     fn fragments_data_in_chunks() {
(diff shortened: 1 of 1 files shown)


DominicBurkart (commented):
## Code Review

The PR is well-structured and the test intent is sound. There are two issues that should be resolved before marking ready-for-review: a significant overlap with PR #106 (already open and not in draft) and a correctness defect in the throttle proptest.

---

### 1. Overlap with PR #106 — needs resolution before merge

PR #106 (`claude/ecstatic-planck-dwPpr-tests`, not in draft) already introduces the following tests for `file_fragmenter.rs`, covering the same invariants as this PR:

| #106 test | #108 near-duplicate |
|---|---|
| `empty_data_is_done_from_creation` | `empty_data_is_immediately_done` |
| `chunk_size_larger_than_data_yields_single_payload` | `chunk_larger_than_data_emits_single_short_chunk` |
| `next_available_time_does_not_mutate_state` | `next_available_time_is_pure` |
| `is_done_only_after_full_consumption` | (covered by `is_done_and_next_available_time_agree`) |
| `concatenated_payloads_roundtrip_original_data` | proptest `chunks_concat_to_original_data` |
| `next_available_time_returns_none_when_done` | (covered by `exhaustion_is_absorbing`) |
| `next_available_time_returns_next_send_when_early` / `_current_when_ready` | `next_available_time_points_at_legal_send_time` |

The three proptest cases (`chunks_concat_to_original_data`, `interval_throttles_payload`, `exhaustion_is_absorbing`) and the module-level doc block are the only genuinely new content relative to #106.

**Options:**
- (a) Rebase this branch on top of the #106 branch and keep only the three proptests and the module doc.
- (b) Close this PR and fold the proptests into #106.

If both PRs merge to `main` as-is, the test suite will contain near-duplicate tests for every `FileFragmenter` invariant.

---

### 2. Throttle proptest — boundary probe produces vacuous assertions

In `interval_throttles_payload`, the inner loop:

```rust
for t in [1u64, interval / 2 + 1, interval - 1] {
    if t == 0 || t >= interval { continue; }
    prop_assert_eq!(f.next_payload(t), None);
}
```

has two problems:

**Problem A — `interval / 2 + 1` can equal `interval` when `interval == 1`.**  
When `interval` is generated as `1`, `interval / 2 + 1 == 1 == interval`, so the `t >= interval` guard skips it. Every element in the array is then skipped and the loop body never executes. The throttle invariant goes unchecked for the smallest (and most interesting) interval value.

**Problem B — `interval - 1` can equal `0` when `interval == 1`.**  
When `interval == 1`, `interval - 1 == 0`, which is also skipped by `t == 0`. Same result: vacuous loop.

The simplest fix is to use a closed-form midpoint that is guaranteed to lie strictly inside `(0, interval)`:

```rust
// Only check the midpoint when the interval is wide enough to have
// a strictly interior integer point.
if interval > 1 {
    let mid = interval / 2;  // 0 < mid < interval because interval > 1
    prop_assert_eq!(
        f.next_payload(mid),
        None,
        "throttle must block at t={mid} < interval={interval}"
    );
}
// One tick before the boundary is always a valid interior point when interval > 1.
if interval > 1 {
    prop_assert_eq!(
        f.next_payload(interval - 1),
        None,
        "throttle must block at t={} < interval={interval}", interval - 1
    );
}
// At the boundary itself the emit must succeed (if data remains).
if !f.is_done() {
    prop_assert!(f.next_payload(interval).is_some());
}
```

Alternatively, restrict `interval in 2u64..=10_000` so `interval == 1` is excluded and the `interval - 1` probe is always valid.

---

### 3. Module-level doc — minor wording issue

The invariant list says `next_available_time` is "idempotent (read-only)". Idempotent means `f(f(x)) == f(x)`, which does not apply to a method that takes a time argument and returns a value. The correct term is **pure** or **read-only** (no observable side effects). The existing tests correctly verify purity, so just the prose needs updating.

The `chunk_size >= 1` precondition is documented but not enforced (no `assert!` or `debug_assert!`). This is acceptable for an example, but worth noting. If enforcement is desired, a `debug_assert!(chunk_size >= 1)` in `FileFragmenter::new` would be the right place.

---

### What is correct

- **Conservation test** (`chunks_concat_to_original_data`): correctly verifies no-loss, no-duplication, no-reordering via reassembly equality. The `chunk.len() <= chunk_size` and `!chunk.is_empty()` assertions are stronger than the equivalent unit test in #106.
- **Exhaustion test** (`exhaustion_is_absorbing`): correctly probes `0`, `1`, `1_000`, `1_000_000`, `u64::MAX/2`, and `u64::MAX`. The absorbing-state framing is clear.
- **Proptest strategies**: `data` up to 200 bytes, `chunk_size` 1–64, `interval` 1–10_000 are all well-bounded. Proptest shrinking will work correctly for these standard strategies.
- **Edge-case unit tests** (`empty_data_is_immediately_done`, `chunk_larger_than_data_emits_single_short_chunk`, `next_available_time_is_pure`, `next_available_time_points_at_legal_send_time`, `is_done_and_next_available_time_agree`): all correct, though duplicated by #106.

---

### Summary verdict

**Not ready for review** in current form due to (1) the overlap issue with #106 that would land duplicate tests on `main`, and (2) the vacuous throttle proptest for `interval == 1`. Fix those two issues — ideally by rebasing to add only the proptests + module doc on top of #106 — and the PR is in good shape.

---

PR #114, opened 2026-05-13
test: cover scheduler/interferer integration path end-to-end

DominicBurkart:
## Summary

The `Scheduler` <-> `InterferenceSource` integration path has many subtle
invariants that the existing suite touches only obliquely. This PR adds
`tests/interferer_node_reply_chain.rs` with 7 focused integration tests
that pin them down end-to-end. Every test is anchored to a specific
invariant in `src/scheduler.rs`; the module docstring documents the
five high-level invariants:

1. Interferer-originated frames trigger `handle_poll_transmit` on every
   non-sender real node, so a node that queues a reply in `on_receive`
   actually transmits in the same scheduler tick.
2. `record_rx` is credited per-node for interferer-originated frames.
3. Capture/collision marking propagates correctly through the scheduler
   when the stronger signal is interferer-originated.
4. Synthetic-ID allocation for many (8) interferers is stable and does
   not produce spurious collisions or airtime drift.
5. `next_poll_time` returning a time inside the interferer's own active
   TX preserves observe ordering (started before completed).

In addition, a determinism test exercises the case where an
interferer-TX completion is coincident with a node wake, and a periodic
burst-jammer test exercises the canonical `poll_inject` ->
`next_poll_time` loop.

## Test plan

- [x] `cargo test --test interferer_node_reply_chain --all-features` -- 7/7 pass locally
- [x] `cargo test --all-features` -- full suite passes (no regressions)
- [x] `cargo clippy --all-features --all-targets -- -D warnings` clean on the new file
- [x] `rustfmt --check tests/interferer_node_reply_chain.rs` clean
- [ ] CI green on push

Notes:
- Only adds a new test file; no changes to library code or other tests.
- Existing unrelated `cargo fmt` drift on `main` is intentionally not
  touched here -- the autoformat CI job already handles that on push.

---
_Generated by [Claude Code](https://claude.ai/code/session_01173Smn5AbVSYT2cT3QKuNN)_

### DIFF
--- a/tests/interferer_node_reply_chain.rs
+++ b/tests/interferer_node_reply_chain.rs
@@ -1,0 +1,667 @@
+//! Integration tests for the scheduler's interferer-originated delivery path.
+//!
+//! These tests pin down behaviour at the boundary between
+//! [`InterferenceSource`] and [`NodeHandle`] that the existing suite touches
+//! only obliquely. Every assertion is anchored to a specific line of
+//! [`src/scheduler.rs`]:
+//!
+//! 1. **Interferer-originated frames reach every node and trigger
+//!    `handle_poll_transmit`.** `deliver_completed_to_nodes` does not special-case
+//!    interferer-originated frames; after `on_receive` it always calls
+//!    `self.handle_poll_transmit(i, time)`. A node that queues a reply on the
+//!    first `on_receive` must therefore actually transmit it in the same
+//!    simulation tick. Without this, adaptive protocols reacting to jammers
+//!    (e.g. ACK-back, listen-before-talk re-arming) would silently break.
+//!
+//! 2. **Per-node `record_rx` attribution for interferer-originated frames.**
+//!    `MetricsCollector::record_rx` is called once per (non-sender) node when a
+//!    non-collided frame completes. The synthetic interferer NodeId
+//!    (`u32::MAX - idx`) never equals a real node id, so the
+//!    `if node_id == sender { continue; }` guard never fires — every real node
+//!    must be credited. The aggregate-rx invariant is verified elsewhere; this
+//!    test pins the *per-node* counter, which protects against a regression
+//!    that double-counts or drops one of the receivers.
+//!
+//! 3. **Capture survives the interferer/node boundary.** If a node TX is
+//!    captured by a stronger overlapping interferer TX, the channel marks the
+//!    node TX collided and the interferer TX captured. The scheduler must then
+//!    record exactly one collision (the node's) and one capture (the
+//!    interferer's), and the interferer's frame must still be delivered to
+//!    every node. The crossover between `Channel::begin_transmission`'s
+//!    capture/collision marking and `deliver_completed_to_nodes`' filtering is
+//!    only exercised in unit tests on the channel, never end-to-end through
+//!    the scheduler with an interferer-originated stronger signal.
+//!
+//! 4. **Many-interferer synthetic-ID allocation is stable.** Adding `N`
+//!    interferers must allocate synthetic ids `u32::MAX, u32::MAX-1, ...,
+//!    u32::MAX-(N-1)` in registration order, and the scheduler must record
+//!    `record_airtime` for every interferer TX. This catches off-by-one or
+//!    re-use bugs in `synthetic_interferer_id`.
+//!
+//! 5. **An interferer's `next_poll_time` may legally schedule a time *before*
+//!    its own TX completes.** The `BinaryHeap` orders events by `time`, so a
+//!    poll scheduled at `t < tx_end` will fire before the `TxComplete` event.
+//!    The interferer must still observe its own `TransmissionStarted` *and*
+//!    `TransmissionCompleted` events, in time order, regardless of when its
+//!    next poll fires.
+
+use std::cell::RefCell;
+use std::rc::Rc;
+
+use theatron::scheduler::{NodeHandle, Scheduler};
+use theatron::time::SimTime;
+use theatron::traits::InterferenceSource;
+use theatron::types::{ChannelEvent, NodeId, RxMetadata, Transmission};
+
+/// (time, sender, kind) — interferer-side observe log entries.
+type ObserveLog = Rc<RefCell<Vec<(SimTime, NodeId, &'static str)>>>;
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+fn make_tx(sf: u8, frequency: u32, duration_us: u64, payload: Vec<u8>, power: i8) -> Transmission {
+    Transmission {
+        payload,
+        sf,
+        bandwidth: 125_000,
+        coding_rate: 5,
+        frequency,
+        duration_us,
+        tx_power_dbm: power,
+    }
+}
+
+/// A node that replies with a queued `Transmission` immediately on its first
+/// `on_receive`. The reply is returned by the very next `poll_transmit`, which
+/// the scheduler invokes inside `deliver_completed_to_nodes` at the same time.
+struct ReplyOnReceiveNode {
+    id: NodeId,
+    received_count: u32,
+    pending_reply: Option<Transmission>,
+    reply_should_fire: bool,
+}
+
+impl NodeHandle for ReplyOnReceiveNode {
+    fn node_id(&self) -> NodeId {
+        self.id
+    }
+    fn on_receive(&mut self, _frame: RxMetadata, _time: SimTime) -> Option<SimTime> {
+        self.received_count += 1;
+        // Arm the reply: it will be drained by the very next poll_transmit
+        // call inside `deliver_completed_to_nodes`.
+        self.reply_should_fire = true;
+        None
+    }
+    fn poll_transmit(&mut self, _time: SimTime) -> Option<Transmission> {
+        if self.reply_should_fire {
+            self.reply_should_fire = false;
+            self.pending_reply.take()
+        } else {
+            None
+        }
+    }
+    fn update(&mut self, _time: SimTime) -> Option<SimTime> {
+        None
+    }
+}
+
+/// A receiver that simply counts how many frames it has received.
+struct CountingReceiver {
+    id: NodeId,
+    received_count: Rc<RefCell<u32>>,
+}
+
+impl NodeHandle for CountingReceiver {
+    fn node_id(&self) -> NodeId {
+        self.id
+    }
+    fn on_receive(&mut self, _frame: RxMetadata, _time: SimTime) -> Option<SimTime> {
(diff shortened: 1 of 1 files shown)


DominicBurkart (commented):
## Review: `tests/interferer_node_reply_chain.rs`

Good foundational integration coverage — the seven tests correctly target distinct invariants and the helpers are clean. Several assertions have subtle mismatches with the actual scheduler/channel behaviour that would cause test failures, and there are a few gaps worth addressing before landing.

---

### Correctness issues

**Test 3 — `total_rx` expected value is wrong**

```rust
assert_eq!(
    sched.metrics.total_rx, 3,
    "strong interferer frame delivered to all 3 real nodes (id=1, 2, 3)"
);
```

The weak node TX is collided, so it delivers nothing. The strong interferer TX is captured and is delivered to every real node whose `node_id` does not equal the sender. The sender is `synthetic_interferer_id(0) = NodeId(u32::MAX)`, which never equals any real `NodeId`, so all three real nodes (1, 2, 3) receive it — but only the two **non-sender** real nodes should receive the interferer's frame *as if it were from a different sender*. Wait — re-reading `deliver_completed_to_nodes`:

```rust
for i in 0..self.nodes.len() {
    let node_id = self.nodes[i].node_id();
    if node_id == sender { continue; }   // sender = NodeId(u32::MAX)
    ...
    self.metrics.record_rx(node_id);
}
```

`sender` is `NodeId(u32::MAX)`, which matches no real node, so all three nodes (1, 2, 3) receive it → `total_rx = 3`. This is actually correct given the implementation.

However the comment in the assertion says `"strong interferer frame delivered to all 3 real nodes (id=1, 2, 3)"` — this is fine.

The assertion that **is** wrong is:

```rust
for id in 1u32..=3 {
    assert_eq!(
        sched.metrics.node_rx_count(NodeId(id)),
        1,
        "node {id} receives the interferer's captured frame exactly once"
    );
}
```

Node 1 is the `OneShotSender`. Its `on_receive` just returns `None` and doesn't increment anything, but `record_rx` is called by the scheduler (not by the node), so `node_rx_count(NodeId(1)) = 1`. This is actually correct too. No bug here — the assertions are right.

**Test 3 — Capture direction in `begin_transmission`**

Looking at the channel logic when the *interferer* starts at `t=10_000` after the *node* started at `t=0`:

```rust
let delta = tx.tx_power_dbm as f32 - active.tx_power_dbm as f32;
// tx     = strong interferer (20 dBm)
// active = weak node      (14 dBm)
// delta  = 20 - 14 = 6
if delta >= self.config.co_channel_rejection_db {   // 6 >= 6 → true
    active.collided = true;   // weak node → collided ✓
    new_captured = true;      // new (interferer) → captured ✓
}
```

Node TX collided, interferer TX captured. However `captured` is set on the new transmission only if `new_captured && !new_collided`. Since `new_collided` is false here, `captured = true`. So `total_collisions = 1`, `total_captures = 1`. ✓

The assertions are correct, but note that the **node TX completes at 80_000** and the **interferer TX at 90_000**, so they resolve in two separate `TxComplete` events. At `t=80_000`: `resolve_at(80_000)` moves the node TX (end=80_000) to `completed`; `drain_completed` processes it → `collided=true` → `record_collision()`. At `t=90_000`: `resolve_at(90_000)` moves the interferer TX → `captured=true` → `record_capture()` → delivered to all 3 nodes. This matches the assertions.

**Test 1 — `total_airtime_us` includes interferer airtime**

The assertion `sched.metrics.total_airtime_us == 40_000 + 30_000` is correct: the scheduler calls `record_airtime(duration)` both in `InterferencePoll` handling (for interferer TX) and in `handle_poll_transmit` (for node TX). Total = 70_000. ✓

**Test 4 — Multiple `TxComplete` events all resolve at the same time**

All 8 interferers are polled at `t=0`, each schedules a `TxComplete` at `t=25_000`. At the first `TxComplete` event, `resolve_at(25_000)` drains **all** active transmissions ending ≤ 25_000 into `completed`, then `drain_completed` delivers all 8 frames to `NodeId(1)` in one go. The remaining 7 `TxComplete` events each call `resolve_at(25_000)` and `drain_completed` on an already-empty active list. This means `node_rx_count(NodeId(1)) = 8`. ✓

---

### Real issue: Test 4 — `total_rx` type mismatch in assertion

```rust
assert_eq!(
    sched.metrics.total_rx, N as u64,   // N is u32 → cast to u64 → 8u64
    ...
);
assert_eq!(
    sched.metrics.node_rx_count(NodeId(1)),
    N as u64,
    ...
);
```

`N` is `u32`, `N as u64` is fine. But `node_rx_count` returns `u64` and `N as u64 = 8`. This compiles and is correct. No issue here.

---

### Real issue: Test 5 — No real nodes registered, but observe loop iterates `self.nodes`

In `deliver_completed_to_nodes`, the loop is over `self.nodes`. Test 5 registers no nodes. This means the interferer's captured frame (if any) is never delivered to any node — but that's fine since the test only checks `observe` callbacks on the interferer itself, not `record_rx`. The test is valid as written.

---

### Gaps and suggestions

**1. `on_receive` return value from `ReplyOnReceiveNode` is unused**

`ReplyOnReceiveNode::on_receive` always returns `None`. The scheduler uses the return value to schedule a follow-up `Wake`. This is intentional (the reply fires via `handle_poll_transmit` inside `deliver_completed_to_nodes`, not via a wake), but the comment says "arm the reply: it will be drained by the very next `poll_transmit` call inside `deliver_completed_to_nodes`". This is correct and the mechanism works — but it relies on the scheduler calling `handle_poll_transmit` *immediately* after `on_receive`, which is exactly invariant 1. The test correctly exercises this, though it would be more explicit to add an assertion that `node_tx_count(NodeId(1)) == 1` is seen at `current_time == 40_000` (the interferer's TX end time), not at a later wake. There is no way to assert on timing after `run()` completes, so the implicit test is acceptable.

**2. Test 3 — `OneShotSender` wakes at `Some(0)` but `poll_transmit` is called from `handle_poll_transmit` during wake, not from `on_receive`**

`OneShotSender::update` returns `None`, and `poll_transmit` drains `self.tx`. This is the correct flow (`Wake` → `update` → `handle_poll_transmit` → `poll_transmit`). Fine.

**3. Missing test: interferer `observe` receives `TransmissionStarted` for *node* TXs**

The current tests verify that nodes receive interferer-originated frames, and that interferers observe their own TX events. But there is no test verifying that interferers observe `TransmissionStarted` events for *node-originated* TXs (from `handle_poll_transmit`). The scheduler does call `interferer.observe(&ch_event, time)` inside `handle_poll_transmit`, and this path is exercised in test 1 (where the reply node TX fires). Adding at least one assertion that the interferer's `observe` log contains the reply node's `TransmissionStarted` entry would close this gap.

**4. `LoggingMidTxInjector::next_poll_time` takes `&self` but mutates shared state via `Rc<RefCell<bool>>`**

This is a valid pattern in Rust (interior mutability), but it means `next_poll_time` is not actually side-effect free — calling it twice would return `None` on the second call regardless of `current_time`. The test relies on this being called exactly once before `peek_time` is needed. This is fine for a test, but worth a comment explaining why the `peeked` flag is `Rc<RefCell<bool>>` rather than a plain `bool` (the reason: `next_poll_time` takes `&self`, not `&mut self`).

**5. Determinism test (test 6) comment overclaims**

```rust
// two events must be processed in FIFO order (by `seq`)
```

The scheduler's `Ord` impl for `ScheduledEvent` uses `other.seq.cmp(&self.seq)` as the tiebreaker inside a `BinaryHeap`. Lower `seq` = earlier insertion = "FIFO" for same-time events. The comment is correct, but the test does not actually *verify* ordering — it only verifies that running the same scenario twice produces the same scalar metrics. If the ordering were wrong in a way that produces different frame-delivery counts, the test would catch it; if the ordering were wrong in a way that is symmetric (same totals, different delivery order), it would not. This is acceptable for a determinism smoke-test, but should be noted in the doc comment.

**6. `PeriodicInjector::next_poll_time` is called *after* `poll_inject` in each `InterferencePoll` event**

Looking at the scheduler:

```rust
EventKind::InterferencePoll { interferer_idx } => {
    ...
    if let Some(tx) = self.interferers[interferer_idx].poll_inject(time) { ... }
    let next = self.interferers[interferer_idx].next_poll_time(time);
    ...
}
```

`poll_inject` decrements `remaining` first, then `next_poll_time` checks `remaining > 0`. So after the last injection (`remaining` goes to 0), `next_poll_time` returns `None` and no more polls are scheduled. Test 7 (`periodic_interferer_delivers_exactly_k_frames`) with `K=5` correctly expects exactly 5 frames. However, `next_poll_time` is called with `current_time` = the *poll* time, not `current_time + interval`. The `PeriodicInjector` adds `interval_us` to `current_time`:

```rust
fn next_poll_time(&self, current_time: SimTime) -> Option<SimTime> {
    if self.remaining > 0 {
        Some(current_time + self.interval_us)
    } else {
        None
    }
}
```

For a periodic jammer starting at 0 with `interval=50_000` and `duration=20_000`: polls at 0, 50_000, 100_000, 150_000, 200_000 (5 total). The 5th poll fires at 200_000 which is within the 1_000_000 simulation window. `total_airtime_us = 5 * 20_000 = 100_000`. ✓

---

### Summary

The tests are logically sound and the assertions match the actual scheduler + channel behaviour as I traced through the code. The main actionable improvements are:

1. **Add an `observe` log assertion in test 1** to verify that interferers observe node-TX `TransmissionStarted` events (currently uncovered).
2. **Add a comment in `LoggingMidTxInjector`** explaining why `peeked` uses interior mutability instead of a `&mut self` flag.
3. **Soften the determinism test comment** from "FIFO order" to "same-time events resolve in insertion order" and note the test's limitation (same totals, not same delivery sequence).
4. **Minor**: `CountingReceiver.received_count` is incremented inside `on_receive` but the counter is never read in tests 2, 3, 4 — only `sched.metrics.node_rx_count` is asserted. Either remove the redundant `Rc<RefCell<u32>>` field (replacing it with a unit struct) or add direct assertions against it to provide an independent cross-check of `MetricsCollector`'s per-node bookkeeping.

None of the above are blockers — the test logic is correct and the coverage is genuinely valuable. Items 1 and 4 are the most worth addressing before merge.


---

DominicBurkart:
**CRON status** — CI is fully green (6/6 checks pass, 100% coverage maintained, branch is clean against main). Code review found no blocking bugs — 4 improvement notes: a coverage gap (no assertion on interferer observing node-originated TX events), a dead `received_count` field on `CountingReceiver`, a missing comment on the `Rc<RefCell<bool>>` interior-mutability workaround in `LoggingMidTxInjector`, and an overclaiming "FIFO order" comment in the determinism test. PR opened for human review.

https://claude.ai/code/session_013pJ49sBBDcA3iazXiMgZav

---

PR #115, opened 2026-05-14
Fix fmt drift and cover channel capture-flag clearing invariant

DominicBurkart:
Maintenance pass across the three workstreams. The repository is already
in very good shape from prior passes, so changes are deliberately small
and focused.

## Architecture
No clear, low-risk non-compliance with `ARCHITECTURE.md` was found. The
source already matches the documented design (event-driven scheduler,
format-agnostic channel, seeded PRNG, interferers as first-class
participants). The minor struct drift (`Transmission` / `RxMetadata`
carry extra physical-layer fields vs. the doc sketches) is consistent
with `ARCHITECTURE.md` being an explicitly-labelled proposal, so it was
left alone rather than churned. No worthwhile LOC/complexity reduction
was identified that wouldn't risk behaviour.

## Testing
Most undertested core behaviour: the capture-effect bookkeeping in
`Channel::begin_transmission`. The capture/collision branches were well
covered *except* the `else` branch that clears a stale `captured` flag
when a third equal-power transmission overlaps an already-captured one.

- Added `captured_flag_cleared_by_later_equal_power_collision` to
  `src/channel.rs`, pinning the invariant that a provisionally captured
  transmission is demoted to `collided` (and un-`captured`) by a later
  equal-power overlap. Without this guarantee a frame that actually
  collided could be reported as captured.

## Clarity
- Re-ran `cargo fmt --all`: the committed tree had drifted from current
  rustfmt output in `examples/lorawan_file_transfer/network_server.rs`
  and `tests/interferer_observe_contract.rs`, which would fail the
  pre-commit `format` hook and the documented
  `cargo fmt --all -- --check` workflow for any local developer.

Local verification: `cargo fmt --all -- --check`, `cargo clippy
--all-features --all-targets -- -D warnings`, and `cargo test
--all-features` all pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01C18s7J23s2b9R64jzfKQAb)_

### DIFF
--- a/examples/lorawan_file_transfer/network_server.rs
+++ b/examples/lorawan_file_transfer/network_server.rs
@@ -56,11 +56,7 @@ mod tests {
     #[test]
     fn on_receive_returns_none() {
         let mut server = NetworkServer::new(NodeId(100));
-        assert!(
-            server
-                .on_receive(make_frame(vec![0x01, 0x02]), 0)
-                .is_none()
-        );
+        assert!(server.on_receive(make_frame(vec![0x01, 0x02]), 0).is_none());
         assert!(server.on_receive(make_frame(vec![0x03]), 1000).is_none());
     }
 
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -544,6 +544,47 @@ mod tests {
         assert!(strong_entry.2, "strongest should be marked captured");
     }
 
+    /// A transmission can be provisionally `captured` by a weaker overlapping
+    /// signal, then later demoted to `collided` (and un-`captured`) when a
+    /// third equal-power signal arrives. This pins the `else` branch of
+    /// `begin_transmission` that clears a stale `captured` flag — otherwise a
+    /// frame that actually collided would still be reported as captured.
+    #[test]
+    fn captured_flag_cleared_by_later_equal_power_collision() {
+        let mut ch = Channel::new();
+        let strong = make_tx_power(7, 868_100_000, 50_000, 20);
+        let weak = make_tx_power(7, 868_100_000, 50_000, 14);
+        // `strong` captures over `weak` (delta = 6 dB == threshold).
+        ch.begin_transmission(NodeId(1), &strong, 0);
+        ch.begin_transmission(NodeId(2), &weak, 10_000);
+        // A third equal-power signal overlaps `strong`: delta = 0 < threshold,
+        // so `strong` must be demoted from captured to collided.
+        let equal = make_tx_power(7, 868_100_000, 50_000, 20);
+        ch.begin_transmission(NodeId(3), &equal, 20_000);
+        ch.resolve_at(70_000);
+        let completed = ch.drain_completed();
+
+        let strong_entry = completed
+            .iter()
+            .find(|(id, _, _, _)| *id == NodeId(1))
+            .unwrap();
+        assert!(
+            strong_entry.1,
+            "strong must be collided after equal-power overlap"
+        );
+        assert!(
+            !strong_entry.2,
+            "strong's stale captured flag must be cleared by the later collision"
+        );
+        // Nothing survives: no entry is both delivered and captured.
+        assert!(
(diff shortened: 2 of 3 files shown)


DominicBurkart (commented):
Reviewed the three-file diff (fmt cleanup + new channel invariant test). No bugs, no logic regressions, no security or PII concerns. All CI checks are green. Detailed notes below.

---

## Fmt changes

`examples/lorawan_file_transfer/network_server.rs` and `tests/interferer_observe_contract.rs` are pure whitespace/formatting: multi-line `assert!` and `Rc<RefCell<…>>` initialisers collapsed to single lines. No logic touched.

---

## New test: `captured_flag_cleared_by_later_equal_power_collision`

### Scenario correctness

The three-step setup is correct and precisely targets the `else` branch of `begin_transmission`:

```
t=0       strong (20 dBm) arrives  — no active peers, stored as collided=false, captured=false
t=10_000  weak   (14 dBm) arrives  — delta = 14−20 = −6 ≤ −threshold
                                      → strong.captured = true
                                      → weak stored as collided=true, captured=false
t=20_000  equal  (20 dBm) arrives  — iterates both active transmissions:
  vs strong: delta = 0, hits else  → strong.collided=true, strong.captured=false (stale flag cleared)
                                      new_collided=true
  vs weak:   delta = 20−14 = +6 ≥ threshold
                                      → weak.collided=true (already), new_captured=true
  equal stored as collided = true (new_collided=true),
             captured = new_captured && !new_collided = false
```

All three transmissions end before `resolve_at(70_000)`, so nothing is left active. The final state matches the assertions exactly.

### Assertions

- `strong_entry.1` (collided=true): correct, set by `equal` arrival.
- `strong_entry.2` (captured=false): correct, this is the key invariant being pinned — the stale flag set by `weak` must be cleared by the later equal-power collision.
- Third assertion (`!any(!collided && captured)`): logically implied by the first two for `strong`, and also checks `weak` and `equal` (both collided). Slightly redundant but harmless and useful as a belt-and-suspenders regression guard.

One nit: the test does not explicitly assert `equal` is also collided. That said, the third assertion covers it indirectly (`equal` is collided so it cannot satisfy `!collided && captured`), and adding an explicit check would be over-specifying. The current level of coverage is appropriate.

### Untested paths in `begin_transmission` claimed by this PR

The PR doc-comment says it pins the `else` branch. The other two branches (`delta >= threshold` and `delta <= -threshold`) are already exercised by `stronger_signal_survives_capture`, `weaker_signal_lost_in_capture`, `just_below_threshold_both_collide`, and `exactly_at_threshold_stronger_survives`. No gaps.

The new test is the first to combine all three branches in a single run (the `equal` arrival triggers `else` for `strong` and `>= threshold` for `weak` simultaneously), which is a meaningful composition.

### Minor observation

The comment on the `weak` arrival reads:

```
// `strong` captures over `weak` (delta = 6 dB == threshold).
```

This is accurate from `strong`'s perspective (strong "wins"), but the delta being evaluated in the loop when `weak` arrives is `weak.tx_power - strong.tx_power = 14 − 20 = −6`, hitting the `<= −threshold` arm (not the `>= threshold` arm). The comment describes the outcome correctly but elides which branch fires. Fine for human readers; just noting it for agents that parse branch coverage from comments.

---

## Summary

LGTM. The fmt diff is clean, and the new test correctly pins the `else`-branch invariant with a well-constructed three-transmission scenario. Assertions are accurate and sufficient.

---

DominicBurkart:
**Cron status (session dVfNL):** Code review posted — fmt changes verified formatting-only, `captured_flag_cleared_by_later_equal_power_collision` correctly pins the demote-to-collided invariant, no bugs or PII found. No owner comments to address; all 6 CI checks green (`Check`, `Test`, `Code Coverage`, `Autoformat`, `Cargo Deny`, `Security Audit`). PR promoted from draft to ready-for-review.

---

